Issue #1, opened 2020-04-24
Transition from Travis CI to GitHub Actions builds + CI, expand with CD support


chrissimpkins:
Steps:

- Add build support #9 
- Add fontbakery CI support #10 
- Add OpenType sanitizer checks #13 
- Add CLDR script/writing system coverage report #16

PR #9, opened 2020-05-20
Add GitHub Actions build CI for static and variable fonts

chrissimpkins:
This PR adds the GH Actions support for reproducible fontmake build CI across the static and variable font build artifact targets.

This also transitions to the system-installed fontmake path in our Makefile definitions so all local installs will need to take place in a venv that is set up with the `make setup` command.

Related: #1

### DIFF
--- a/.github/workflows/static-artifact-ci.yml
+++ b/.github/workflows/static-artifact-ci.yml
@@ -1,0 +1,52 @@
+name: Static CI
+
+on: [push, pull_request]
+
+jobs:
+  static-font-ci-job:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [3.8]
+    name: Google Sans static font build CI
+    steps:
+      - name: Configuration
+        id: config
+        uses: f-actions/font-setup@master
+        with:
+          projectname: "Google Sans"
+          sourcepath: "src"
+          buildpath: "build/GoogleSans/static/expert/*.ttf" # static font build path
+          readmepath: "README.md"
+          changelogpath: "CHANGELOG.md"
+          dependpath: "requirements.txt"
+          py-version: ${{ matrix.python-version }}
+          buildcommand: "make gs-static" # static font make target
+      - name: Action environment variable definitions
+        run: |
+          echo "projectname: ${{ steps.config.outputs.projectname }}"
+          echo "sourcepath: ${{ steps.config.outputs.sourcepath }}"
+          echo "buildpath: ${{ steps.config.outputs.buildpath }}"
+          echo "readmepath: ${{ steps.config.outputs.readmepath }}"
+          echo "changelogpath: ${{ steps.config.outputs.changelogpath }}"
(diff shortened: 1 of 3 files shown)


anthrotype:
> transitions to the system-installed fontmake path in our Makefile definitions

why? Calling the fontmake with explicit relative path to the venv/bin was a feature! At least I think it's preferable than requiring the user to source the activate script.

anthrotype:
it's also risky because it may inadvertently end up using an unrelated fontmake that happens to be on the PATH at the time of running the makefile. Anyway, that's just my opinion, feel free to ignore it.

chrissimpkins:
Agree.  Unless I set up a venv in the remote CI I can't use the same path and haven't been able to come up with a way to switch it based on CI env vars...

This new approach is intended to move all teams away from local builds though. Our GH Actions CI will build artifacts and upload them to a download link so that they only need to push commits to design sources and click to download files as soon as the build is complete.  There are a number of teams that are running into issues with Python tooling installs/venv management, etc.  I am going to take it out of their hands and not depend on anyone to have an active venv environment to build the project.

anthrotype:
I don't understand.. If it works locally, how come should it not also work on a CI?

chrissimpkins:
Because I don't build a venv on the CI.  The path pointed to a .venv directory in the root that doesn't exist.  We build with system installed paths of everything in the runner env.

chrissimpkins:
The runner env is the venv :)

anthrotype:
> not depend on anyone to have an active venv environment

that was exactly the purpose of bootstrapping a venv from the Makefile and calling the venv's bin scripts with explicit paths.

anthrotype:
> The runner env is the venv :)

that's an implementation detail :) You can always make another venv from a venv recursively at infinitum...

anthrotype:
nevermind, you do you what you think it's best 👍

chrissimpkins:
No, I definitely appreciate the feedback.  The root issue though is that I *don't want* anyone to build locally.  The intent here is that they will not ever do this.  The entire CI process will be set up to support remote builds that they can simply download so that there is never confusion about dependency management and venv's.  There is never a guarantee that a design team bumps their own local requirements.txt dependencies when we escalate the definitions and there isn't a check in place to confirm these dependencies in their venv's.  This requires them to follow the commits on the repository and they don't do that.  At the moment they work in forks that get dozens/hundreds of commits behind the upstream.  Since the decision was made to try to merge most of the scripts into the same family, I am going to try to streamline the development into the same repository.  Remains to be seen how this will work but there have been problems with the previous approach.

I am interested in your thoughts about the full approach.  This is just the beginning. I should have everything in place today and will give you the tour of what I have.

chrissimpkins:
And we can definitely convert to building a venv before we perform the builds so that there is improved local and remote support.  It is a couple of lines in the Makefiles and another step in the env set up here.  Let's discuss once the full CI is in place.

PR #10, opened 2020-05-20
Add GH Actions fontbakery testing of static and variable font artifacts

chrissimpkins:
Related #1 

This PR adds fontbakery testing with custom Google Sans check profiles.  This adds project-specific tests to the upstream universal profile that is executed on all build artifacts.

### DIFF
--- a/.github/workflows/static-artifact-ci.yml
+++ b/.github/workflows/static-artifact-ci.yml
@@ -22,15 +22,6 @@ jobs:
           dependpath: "requirements.txt"
           py-version: ${{ matrix.python-version }}
           buildcommand: "make gs-static" # static font make target
-      - name: Action environment variable definitions
-        run: |
-          echo "projectname: ${{ steps.config.outputs.projectname }}"
-          echo "sourcepath: ${{ steps.config.outputs.sourcepath }}"
-          echo "buildpath: ${{ steps.config.outputs.buildpath }}"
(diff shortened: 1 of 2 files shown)


chrissimpkins:
I have a bit of fontbakery cleanup to do with new build targets.

Recent versions of fontbakery are now failing the default VF STAT tables coming from fontmake. Will deactivate the test on these builds until I implement the STAT tables on these builds.  This is not a high priority target until the Fonts API supports our opsz axis range. 

The default builds are failing on the fea tag observed vs. expected check.  This is not an issue with the builds.  The definitions need to be updated for the new "Default" build targets requested by Jonathan Lee and Damien Correll.  We are removing `aalt` supported alternates and all alternate outlines from the builds and that fea needs to be removed from the expected set for tests on these targets.

chrissimpkins:
https://github.com/googlefonts/googlesans/pull/10/commits/8f93d8ed2510e0be373f81848a59967ca927f156 removes the expert VF fontbakery checks.

Opened #11 to track the STAT table fix and reactivation of these tests when this is in place.

chrissimpkins:
https://github.com/googlefonts/googlesans/pull/10/commits/e0720748cc259ea2d9c4e291e45f85bc048fd9b1 removes the "Default" static build target fontbakery checks. These are not our current production builds.  This transition will occur as of GS v4.

Added #12 to track that this is reactivated as part of the transition to the new default build targets.

PR #14, opened 2020-05-20
Add remote WIP font build push + download support through GH Actions

chrissimpkins:
This PR adds support for uploads of static and variable font build artifacts when commits are pushed to `master`, `build-*`, or `build/*` git branches.

The artifacts are compiled with the same reproducible build pipeline that is checked by our CI tests and pushed for download through the GitHub Actions UI.

This will support WIP testing of font files during development.  These push workflows are intentionally separate from the CI test workflows so that builds + pushes occur even when CI tests fail as binaries may need to be inspected as part of CI failure diagnostics. 

These artifacts are stored for 90 days. They are meant to be temporary builds for testing support.  Releases will be pushed as permanent artifacts using GitHub releases when a version formatted git tag is pushed to the remote.  Release artifact support will be added in a separate PR.

With these changes, designers will be able to push design source updates to a `build-*` or `build/*` branch and click to download test fonts without a local build environment requirement.

### DIFF
--- a/.github/workflows/upload-artifacts.yml
+++ b/.github/workflows/upload-artifacts.yml
@@ -1,0 +1,93 @@
+name: Artifact Uploads
+on:
+  push:
+    # Sequence of patterns matched against refs/heads
+    branches:
+      - master # Push events on master branch
+      - "build/**" # Push events to branches matching refs/heads/build/[ANYTHING]
+      - "build-*" # Push events to branches matching refs/heads/build-[ANYTHING]
+
+jobs:
+  static-artifact-uploads:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [3.8]
+    name: Upload static artifacts
+    steps:
+      - name: Configuration
+        id: config
+        uses: f-actions/font-setup@master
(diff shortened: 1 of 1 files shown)


chrissimpkins:
@anthrotype Cosimo, here is what I have in mind for the remote build workflow to support the design teams.  They push design source updates and download their fonts from the GitHub UI using our reproducible build pipeline without the need to manage any of the Py dependencies / venv issues locally.  

See https://github.com/googlefonts/googlesans/actions/runs/110761104 for an example of the UI access to the artifacts.

anthrotype:
Looks really good! What's that font-setup action doing?

chrissimpkins:
It is a custom GH Action that configures the github env vars for use further down in the configuration file.  I am designing the build workflow so that it can (hopefully) be used in other projects.  It wraps around existing build scripting (shell, Makefiles, Py scripts) and I am beginning to add elements to support common CI tasks.  My goal is to be able to make a set of customizable template configuration files so that new projects fill out the metadata in that font-setup section, push the yaml files to GH, and it takes care of the rest.

PR #15, opened 2020-05-20
Add build time UPM scaling support to variable font artifact pipeline

### LINKED ISSUE #32: Increase UPM in variable fonts to 2000 at build time
Request from Dave Crossland to scale from 1000 UPM to 2000 UPM at build time in order to limit integer rounding effects on design axis interpolations

Implementation: #15 

---

Transferred from https://github.com/Colophon-Foundry/google-sans/issues/44

### LINKED ISSUE #33: Update fontbakery expected metrics values for 2000 UPM scaled variable fonts
We need to scale the expected values in the custom fontbakery checks so that they match the scaled values in the binaries.

Related #15

### LINKED ISSUE #48: Fix UPM scaling script to support all design source layers
There is an opsz axis interpolation issue in the builds as of the 2000 UPM scaling work @ https://github.com/googlefonts/googlesans/commit/e9d0308c9c1dd5b5966b6b88fde190eab74b9aab:

VF-figures

cc @Colophon-Foundry

chrissimpkins:
This PR adds a [UPM scaling script](https://gist.github.com/anthrotype/62d0bfe1d38b8f11a199bd3b66574bcc) that works at the intermediate UFO stage of the fontmake variable font build process.  

#### TODO

- [ ] update fontbakery metrics value checks for variable fonts (must be scaled by factor of 2 with these changes)
- [x] check file size difference b/w 1000 UPM and 2000 UPM VF builds (~2% increase in size with 2000 UPM)
- [x] fix upm scaling script so that it scales values across all layers in the design (b289377)
- [ ] refactor static and VF make targets to eliminate duplication/take advantage of steps that can be parallelized in the build process with `make -j`

Closes #32 
Closes #33 
Fixes #48

### DIFF
--- a/source/Makefile
+++ b/source/Makefile
@@ -22,6 +22,13 @@ VENV_DIR=.venv
 PYTHON3_EXE=python3
 
 # --------------------------------------------------------
+#  Variable font UPM value
+#    NOTE: used for scaling of variable font UFO sources
+#          at build time
+# --------------------------------------------------------
+UPM=2000
+
+# --------------------------------------------------------
 #  fontmake compiler executable
 #    NOTE: this builds from a venv install to support
 #          pinned dependency versions
@@ -78,21 +85,27 @@ UPRIGHT_DESIGNSPACE_PATH=$(MASTER_UFO_DIR)/$(UPRIGHT_DESIGNSPACE)
 ITALIC_DESIGNSPACE_PATH=$(MASTER_UFO_DIR)/$(ITALIC_DESIGNSPACE)
 
 # --------------------------------------------------------
-#  Master and instance intermediate UFO file paths
+#  Master intermediate UFO file paths
 # --------------------------------------------------------
-UFO_GS_REGULAR=$(MASTER_UFO_DIR)/GoogleSans-Regular.ufo
-UFO_GS_ITALIC=$(MASTER_UFO_DIR)/GoogleSans-Italic.ufo
-UFO_GS_MEDIUM=$(INSTANCE_UFO_DIR)/GoogleSans-Medium.ufo
-UFO_GS_MEDIUMITALIC=$(INSTANCE_UFO_DIR)/GoogleSans-MediumItalic.ufo
-UFO_GS_BOLD=$(MASTER_UFO_DIR)/GoogleSans-Bold.ufo
-UFO_GS_BOLDITALIC=$(MASTER_UFO_DIR)/GoogleSans-BoldItalic.ufo
+UFO_GS_REGULAR=$(MASTER_UFO_DIR)/GoogleSans-SansRegular.ufo
+UFO_GS_ITALIC=$(MASTER_UFO_DIR)/GoogleSans-SansItalic.ufo
+UFO_GS_BOLD=$(MASTER_UFO_DIR)/GoogleSans-SansBold.ufo
+UFO_GS_BOLDITALIC=$(MASTER_UFO_DIR)/GoogleSans-SansBoldItalic.ufo
 
 UFO_GS_TEXT_REGULAR=$(MASTER_UFO_DIR)/GoogleSans-TextRegular.ufo
 UFO_GS_TEXT_ITALIC=$(MASTER_UFO_DIR)/GoogleSans-TextItalic.ufo
+UFO_GS_TEXT_BOLD=$(MASTER_UFO_DIR)/GoogleSans-TextBold.ufo
+UFO_GS_TEXT_BOLDITALIC=$(MASTER_UFO_DIR)/GoogleSans-BoldTextItalic.ufo
+
+# --------------------------------------------------------
+#  Instance intermediate UFO file paths
+# --------------------------------------------------------
+
+UFO_GS_MEDIUM=$(INSTANCE_UFO_DIR)/GoogleSans-Medium.ufo
+UFO_GS_MEDIUMITALIC=$(INSTANCE_UFO_DIR)/GoogleSans-MediumItalic.ufo
+
 UFO_GS_TEXT_MEDIUM=$(INSTANCE_UFO_DIR)/GoogleSans-TextMedium.ufo
 UFO_GS_TEXT_MEDIUMITALIC=$(INSTANCE_UFO_DIR)/GoogleSans-TextMediumItalic.ufo
-UFO_GS_TEXT_BOLD=$(MASTER_UFO_DIR)/GoogleSans-TextBold.ufo
-UFO_GS_TEXT_BOLDITALIC=$(MASTER_UFO_DIR)/GoogleSans-TextBoldItalic.ufo
 
 # --------------------------------------------------------
 #  TARGETS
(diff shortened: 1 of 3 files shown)


chrissimpkins:
@anthrotype This is the initial attempt at scaling to 2000 UPM without the addition of make targets to support parallel builds at this stage.  I want to make sure that I have this right before I dive into the parallel builds.  Mind having a look at this?  

If I understood our conversation correctly, the fontmake build was glyphs -> designspace + UFO with interpolated instance UFO's, followed by instance UFO path -> ttf.  Is this for static instances rather than the VF's?  I build to designspace x 2 (=upright and italic) + 4-master UFO x 2, 6-instance UFO x 2, and can only get to the VF's by pointing fontmake to the designspace in the `master_ufo` dir correct?  Or do I include all four master or six instance UFO sources as args for the upright VF build and the same with the italics VF build using the italics sources? It looks like the `-u` option accepts more than one UFO path.

When I add the scaling of the *master* UFO's to the likely incorrect current approach, I see this new error:

```
fontmake -m ../build/GoogleSans/master_ufo/GoogleSans.designspace -o variable --output-dir ../build/GoogleSans/variable/expert
INFO:fontmake.font_project:Building variable font ../build/GoogleSans/variable/expert/GoogleSans-VF.ttf
INFO:ufo2ft:Pre-processing glyphs
ERROR:cu2qu.ufo:Glyphs named 'section' have incompatible segment types:
   3: ('line', 'curve', 'line', 'curve')
   9: ('line', 'curve', 'line', 'curve')
  19: ('line', 'curve', 'line', 'curve')
  27: ('line', 'curve', 'line', 'curve')
ERROR:cu2qu.ufo:Glyphs named 'Uogonek' have incompatible segment types:
  16: ('line', 'line', 'curve', 'curve')
ERROR:cu2qu.ufo:Glyphs named 'eogonek' have incompatible segment types:
  15: ('line', 'line', 'curve', 'curve')
fontmake: Error: In '../build/GoogleSans/master_ufo/GoogleSans.designspace': Generating fonts from Designspace failed: fonts contains incompatible glyphs: 'Uogonek', 'eogonek', 'section'
```

I'm guessing that this is a build approach issue and not a scaling problem.

Logs are in https://github.com/googlefonts/googlesans/runs/694564327?check_suite_focus=true.

source/Makefile
```diff
@@ -125,8 +138,27 @@ gs-static: $(FONT_BUILD_DIR)
 
 
 gs-vf: $(FONT_BUILD_DIR)
-	$(FONTMAKE_EXE) -g $(UPRIGHT_SOURCE) -o variable --output-dir $(EXPERT_VARIABLE_BUILD_DIR) --master-dir $(MASTER_UFO_DIR) --instance-dir $(INSTANCE_UFO_DIR)
-	$(FONTMAKE_EXE) -g $(ITALIC_SOURCE) -o variable --output-dir $(EXPERT_VARIABLE_BUILD_DIR) --master-dir $(MASTER_UFO_DIR) --instance-dir $(INSTANCE_UFO_DIR)
+	@echo "================================"
+	@echo " Build variable artifacts..."
+	@echo "================================"
+	# Build to intermediate UFO + designspace files
+	$(FONTMAKE_EXE) -g $(UPRIGHT_SOURCE) --interpolate -o ufo --output-dir $(EXPERT_VARIABLE_BUILD_DIR) --master-dir $(MASTER_UFO_DIR) --instance-dir $(INSTANCE_UFO_DIR)
```
anthrotype: for the VF, you only want to build the master UFOs + designspace, not also the instances. Therefore you don't want to pass --interpolate option (which will produce the instances as well).

```suggestion
	$(FONTMAKE_EXE) -g $(UPRIGHT_SOURCE) -o ufo --output-dir $(EXPERT_VARIABLE_BUILD_DIR) --master-dir $(MASTER_UFO_DIR) --instance-dir $(INSTANCE_UFO_DIR)
```

anthrotype (commented):
unfortunately fontMath seems to not guarantee to maintain interpolation compatibility of the masters after applying the scale operation. It does some massaging to the glyph contour points such as normalizing all incoming lines to curves and then afterwards filtering out unnecessary contour points. It was originally designed to help with making static interpolated UFO instances (via mutatorMath). If we want to pursue this approach of scaling the UFO masters we will have to apply the 2x transformation ourselves (e.g. using a TransformPointPen).
I suggest that you first confirm that there is a real problem to solve here with this build-time scaling of UPEM, and then we can see how to solve it.

For gauging the file size increase after scaling you can compare the static instances before/after scaled by the scale_ufo.py script.

chrissimpkins:
Thanks Cosimo.  This is the proper approach to build the VF?

anthrotype:
> This is the proper approach to build the VF?

yes, you first build the master UFOs + designspace, then scale the masters then proceed to build the VF from the Designspace file.

I updated the scale_ufo.py script to no longer use fontMath. It was creating more problems than it was solving.
Please upgrade to ufo2ft 2.14.0 since I have fixed another issue related to the TransformationsFilter.

https://gist.github.com/anthrotype/62d0bfe1d38b8f11a199bd3b66574bcc

chrissimpkins:
> > This is the proper approach to build the VF?
> 
> yes, you first build the master UFOs + designspace, then scale the masters then proceed to build the VF from the Designspace file.
> 
> I updated the scale_ufo.py script to no longer use fontMath. It was creating more problems than it was solving.
> Please upgrade to ufo2ft 2.14.0 since I have fixed another issue related to the TransformationsFilter.
> 
> https://gist.github.com/anthrotype/62d0bfe1d38b8f11a199bd3b66574bcc

I'll update this today and push a new commit.  Thanks Cosimo!

chrissimpkins:
> I suggest that you first confirm that there is a real problem to solve here with this build-time scaling of UPEM, and then we can see how to solve it.

Agree.  See the email from David Berlow that I sent earlier today.  That is the only example that I have to date.  Our Latin, CJK, Indic, and Thai design teams are looking into it.

chrissimpkins:
> Please upgrade to ufo2ft 2.14.0 since I have fixed another issue related to the TransformationsFilter.

> I updated the scale_ufo.py script to no longer use fontMath

> Therefore you don't want to pass --interpolate option

All addressed as of [https://github.com/googlefonts/googlesans/pull/15/commits/a5299e4d66a3b5e89e11c0e693ac3b12a2cdbfab ](https://github.com/googlefonts/googlesans/pull/15/commits/a5299e4d66a3b5e89e11c0e693ac3b12a2cdbfab)

Let's see what happens with this build

chrissimpkins:
Builds work.  And now all of my pinned metrics attributes checks fail ha!  We are definitely scaling by a factor of two.  One more FB check to update...

chrissimpkins:
I'll have a look at the fonts today and push data about the file size change with these changes.  Will continue to investigate the problem so that we have a test case to determine if this is a fix.

source/Makefile
```diff
@@ -125,8 +138,27 @@ gs-static: $(FONT_BUILD_DIR)
...
 
 gs-vf: $(FONT_BUILD_DIR)
-	$(FONTMAKE_EXE) -g $(UPRIGHT_SOURCE) -o variable --output-dir $(EXPERT_VARIABLE_BUILD_DIR) --master-dir $(MASTER_UFO_DIR) --instance-dir $(INSTANCE_UFO_DIR)
-	$(FONTMAKE_EXE) -g $(ITALIC_SOURCE) -o variable --output-dir $(EXPERT_VARIABLE_BUILD_DIR) --master-dir $(MASTER_UFO_DIR) --instance-dir $(INSTANCE_UFO_DIR)
+	@echo "================================"
+	@echo " Build variable artifacts..."
+	@echo "================================"
+	# Build to intermediate UFO + designspace files
+	$(FONTMAKE_EXE) -g $(UPRIGHT_SOURCE) -o ufo --output-dir $(EXPERT_VARIABLE_BUILD_DIR) --master-dir $(MASTER_UFO_DIR) --instance-dir $(INSTANCE_UFO_DIR)
+	$(FONTMAKE_EXE) -g $(ITALIC_SOURCE) --interpolate -o ufo --output-dir $(EXPERT_VARIABLE_BUILD_DIR) --master-dir $(MASTER_UFO_DIR) --instance-dir $(INSTANCE_UFO_DIR)
```
anthrotype:
```suggestion
	$(FONTMAKE_EXE) -g $(ITALIC_SOURCE) -o ufo --output-dir $(EXPERT_VARIABLE_BUILD_DIR) --master-dir $(MASTER_UFO_DIR) --instance-dir $(INSTANCE_UFO_DIR)
```

chrissimpkins: Ah sorry.  I accepted the change above and didn't change the other one.  Adding this commit.

anthrotype: I suggest you split this designspace building step into its own make task, so you can reuse it for building the static instances. It is shared by both VF and static builds.

chrissimpkins:
It is a ~ 2% increase in file size to scale to 2000 UPM from 1000 UPM.  

For the 2-axis builds:

### 2000 UPM 

- upright: 581kb
- italic:  605kb


### 1000 UPM

- upright: 570kb
- italic:  592kb


Data are similar for 1-axis partial instances (variable wght axis only) when you consider the full design space.

anthrotype:
> It is a ~ 2% increase in file size to scale to 2000 UPM from 1000 UPM.

/cc @rsheeter

anthrotype:
@chrissimpkins sorry, I just realized that the scale_ufo.py script was not also updating the advance width.. please use this fixed copy, thanks https://gist.github.com/anthrotype/62d0bfe1d38b8f11a199bd3b66574bcc/revisions

chrissimpkins:
> @chrissimpkins sorry, I just realized that the scale_ufo.py script was not also updating the advance width.. please use this fixed copy, thanks https://gist.github.com/anthrotype/62d0bfe1d38b8f11a199bd3b66574bcc/revisions

No worries Cosimo! [https://github.com/googlefonts/googlesans/pull/15/commits/80a8274232c1d5eccf95ca31f7f06bf88c415f3c ](https://github.com/googlefonts/googlesans/pull/15/commits/80a8274232c1d5eccf95ca31f7f06bf88c415f3c)

Don't hesitate to push to this repository btw.  We moved it with the intent to make the work / builds more accessible and use the threads to support discussions about the project across internal / external SH's.  I'll be inviting everyone to the repo next week once the set up and transition is complete.

chrissimpkins:
Rebased on master @ https://github.com/googlefonts/googlesans/commit/e9d0308c9c1dd5b5966b6b88fde190eab74b9aab after merge of https://github.com/googlefonts/googlesans/pull/46

chrissimpkins:
Possibly related to the scaling here? https://github.com/googlefonts/googlesans/issues/48

This occurs in our 2000 UPM scaled VF builds and only appears to affect the figures with brace layers.  The brace layer definitions differ between the italics and upright sources (more figure glyphs use them in italics)

anthrotype:
I have a fix, but I think I just pushed it to the wrong branch "build-2000". I see this one is called "chrissimpkins-upm-scaling"...
Which one should I use?

anthrotype:
Anyway, this is the commit
https://github.com/googlefonts/googlesans/commit/14a77e93097e4911d37f9c7057f72b7e0cf14bfe

you can cherry pick wherever you like

chrissimpkins:
No problem at all.  The `build-*` branches will build artifacts.  This is good.  I'll have a look at the builds and cherry pick into the PR branch once I confirm the fix.  Thank you!

chrissimpkins:
Confirmed fix of #48.  Thank you very much Cosimo!

chrissimpkins:
This will not be implemented.  See https://github.com/googlefonts/googlesans/issues/32#issuecomment-647537804

chrissimpkins:
Keeping this branch for now so that we have Cosimo's script in case the need arises again.

chrissimpkins:
Closing as will not fix. Keeping this branch for now for reference purposes.

madig:
Can we delete the branch now?

PR #60, opened 2020-06-18
Min Opsz restructure

Colophon-Foundry:
- Remove Brace layers `{15}`
- Update Bracket layers to `17.5` from `16` (Majority)
- Shift Master position to `17` from `14`

The shift to 17.5 is needed to initiate the glyph swap in forms (G y etc).

@chrissimpkins The build will still fail I believe because of the shift in naming for the brace/bracket layers.

Below is a quick snippet for spotting all brace/bracket layers (Just change the if statement from `[` to `{`)

``` python
import GlyphsApp

Font = Glyphs.font
editString = ""
for thisGlyph in Font.glyphs:
	if "[" in "".join([ l.name for l in thisGlyph.layers ]):
		editString += "/" + thisGlyph.name
		print thisGlyph.name
		print l.name
```

### DIFF
--- a/source/GoogleSans/GoogleSans-Italic.glyphs
+++ b/source/GoogleSans/GoogleSans-Italic.glyphs
@@ -1,5 +1,5 @@
 {
-.appVersion = "1286";
+.appVersion = "1237";
 classes = (
 {
 code = "zero one two three four five six seven eight nine";
@@ -3206,7 +3206,7 @@ position = "{91, 960}";
 );
 id = "15F39A2A-7598-44C5-A4AA-18A6AD645EAA";
 italicAngle = 10;
-weightValue = 14;
+weightValue = 17;
 widthValue = 380;
 xHeight = 526;
 },
@@ -3301,7 +3301,7 @@ verticalStems = (
 95
 );
 weight = Bold;
-weightValue = 14;
+weightValue = 17;
 widthValue = 734;
 xHeight = 526;
 },
@@ -9104,7 +9104,7 @@ unicode = 0046;
 },
 {
 glyphname = G;
-lastChange = "2020-01-22 16:45:12 +0000";
+lastChange = "2020-06-17 16:50:16 +0000";
 layers = (
 {
 anchors = (
@@ -9370,7 +9370,7 @@ position = "{476, 716}";
 );
 associatedMasterId = "15F39A2A-7598-44C5-A4AA-18A6AD645EAA";
 layerId = "4EA8DA53-A620-4DD1-9424-B60B28B7E4D3";
-name = "[16]";
+name = "[17.5]";
 paths = (
 {
 closed = 1;
@@ -9477,7 +9477,7 @@ position = "{493, 716}";
 );
 associatedMasterId = "3E1A77A5-2AAF-4B97-842E-5802590C9619";
 layerId = "FC80B4FC-6917-4D48-AD4D-84FB370372BA";
-name = "[16]";
+name = "[17.5]";
 paths = (
 {
 closed = 1;
@@ -9577,7 +9577,7 @@ nodes = (
 );
 };
 layerId = "418B8A48-8E35-4E59-8A7E-F3FF83C640EE";
-name = "[16]";
+name = "[17.5]";
 paths = (
 {
 closed = 1;
@@ -9677,7 +9677,7 @@ nodes = (
 );
 };
 layerId = "9E059F69-376B-48B9-A0F0-0B2EC2DC38CE";
-name = "[16]";
+name = "[17.5]";
 paths = (
 {
 closed = 1;
@@ -9725,7 +9725,7 @@ unicode = 0047;
 },
 {
 glyphname = Gbreve;
-lastChange = "2020-01-22 16:45:12 +0000";
+lastChange = "2020-06-17 16:50:26 +0000";
 layers = (
 {
 components = (
@@ -9791,7 +9791,7 @@ transform = "{1, 0, 0, 1, 227, 0}";
 }
 );
 layerId = "5AE0E551-FEA0-4610-BA33-C7274B9F758E";
-name = "[16]";
+name = "[17.5]";
 width = 748;
 },
 {
@@ -9806,7 +9806,7 @@ transform = "{1, 0, 0, 1, 236, 0}";
 }
 );
 layerId = "5D89527F-D23D-4D99-81BA-E1C9D163ABE6";
-name = "[16]";
+name = "[17.5]";
 width = 794;
 },
 {
@@ -9821,7 +9821,7 @@ transform = "{1, 0, 0, 1, 235, 0}";
 }
 );
 layerId = "42BD0172-88E6-4E98-84D2-4CBE9F6CE3FB";
-name = "[16]";
+name = "[17.5]";
 width = 814;
 },
 {
@@ -9836,7 +9836,7 @@ transform = "{1, 0, 0, 1, 230, 0}";
 }
 );
 layerId = "05F9A34A-FB17-4C34-8CDB-3E2B379EE0F4";
-name = "[16]";
+name = "[17.5]";
 width = 826;
 }
 );
@@ -9846,7 +9846,7 @@ unicode = 011E;
 },
 {
 glyphname = Gcircumflex;
-lastChange = "2020-01-22 16:45:12 +0000";
+lastChange = "2020-06-17 16:50:34 +0000";
 layers = (
 {
 components = (
@@ -9912,7 +9912,7 @@ transform = "{1, 0, 0, 1, 229, 0}";
 }
 );
 layerId = "A083780D-FACF-4D01-8905-64DA8997F4C9";
-name = "[16]";
+name = "[17.5]";
 width = 748;
 },
 {
@@ -9927,7 +9927,7 @@ transform = "{1, 0, 0, 1, 215, 0}";
 }
 );
 layerId = "DEDF4123-CC98-4A5D-BEDF-4914AFE75CF4";
-name = "[16]";
+name = "[17.5]";
 width = 794;
 },
 {
@@ -9942,7 +9942,7 @@ transform = "{1, 0, 0, 1, 239, 0}";
 }
 );
 layerId = "788F724F-2D47-47A0-8952-048ED6C4A066";
-name = "[16]";
+name = "[17.5]";
 width = 814;
 },
 {
@@ -9957,7 +9957,7 @@ transform = "{1, 0, 0, 1, 211, 0}";
 }
 );
 layerId = "666C8A43-722B-4B18-8683-C89776D5BA9E";
-name = "[16]";
+name = "[17.5]";
 width = 826;
 }
 );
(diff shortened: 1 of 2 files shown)


chrissimpkins:
Tyvm Edd!

chrissimpkins:
@alexeiva @gorjious @razmadzekoko  note the opsz axis definition changes here.  We will need these same changes in all scripts.  Please let me know if you have any questions.

chrissimpkins:
We don't get the expected snapping behavior with forms like cap `G` (no spur expected in max opsz designs) and lower `y` (no tail expected in max opsz designs) with the changes as of https://github.com/googlefonts/googlesans/pull/60/commits/809f85dc015c9666229fc2b4558daf959b248820

chrissimpkins:
2020-06-22_12-05-02-crunch

chrissimpkins:
It looks like we need to push the bracket layers that were previously defined at 16 to **18** instead of **17.5**.  Those bracket layers are the max opsz designs and they appear to function appropriately when defined at whole integer values (top line for uprights), but not (in Axis-Praxis or Font Goggles) when defined at 17.5 (bottom line italics was not modified).   If I define at 17 instead, they remain pinned on the min opsz forms and never transition as seen in the 17.5 definitions.

GS-opsz

cc @anthrotype so that you are aware.  Note cap `G` and lower `y` form transitions.  I'm not sure that it is ever necessary to define at intermediate values between integer steps here but it does not appear to work with fontmake pipeline based builds.  IMO we don't need it at an intermediate value.  Let's see what Edd thinks (@Colophon-Foundry).

chrissimpkins:
@Colophon-Foundry 

Edd, I updated all of the bracket layers automatically with the following updates to your script using the Glyphs Macro editor:

```python
import GlyphsApp

PRE_BRACKET_SIZE = "17.5"
POST_BRACKET_SIZE = "18"

Font = Glyphs.font
try:
    for thisGlyph in Font.glyphs:
    	for layer in thisGlyph.layers:
    		if "[" in layer.name and PRE_BRACKET_SIZE in layer.name:
    		    layer.name = layer.name.replace(PRE_BRACKET_SIZE, POST_BRACKET_SIZE)
    		    print("{} in {} updated to {}".format(thisGlyph.name, layer.master, POST_BRACKET_SIZE))
except TypeError as e:
	print("ERROR: {}".format(e))
	
Font.save()
```

The output from the script runs on the upright and italics source files is attached in the text files below so that you can review the glyphs that were modified.  Mind having a look at the variable fonts that are built at the commit that I pushed here ->  https://github.com/googlefonts/googlesans/suites/827675683/artifacts/9204584

Please let me know if you agree with these changes.

[GS-opsz-18-upright.txt](https://github.com/googlefonts/googlesans/files/4815171/GS-opsz-18-upright.txt)
[GS-opsz-18-italics.txt](https://github.com/googlefonts/googlesans/files/4815172/GS-opsz-18-italics.txt)

---

@razmadzekoko @gorjious You should be able to use the script above to automatically edit the bracket layers in your Glyphs source files.  You will need to change the line:

```python
PRE_BRACKET_SIZE = "17.5"
```
to:

```python
PRE_BRACKET_SIZE = "16"
```

or whatever value you are using for the number inside your bracket layer name definitions. e.g., `[16]` would be defined as `"16"` in this script.

chrissimpkins:
Sources were pushed to this branch https://github.com/googlefonts/googlesans/tree/Latin-18-opsz and are current with the changes to the bracket layers as of commit https://github.com/googlefonts/googlesans/commit/3b126f740de1738d9106aad0c1e3f6c24257ade0

chrissimpkins:
I added a PR of those changes against the Latin branch so that you can view the diff across the source files relative to where you are here.

https://github.com/googlefonts/googlesans/pull/61

Colophon-Foundry:
@chrissimpkins Thanks for looking into this. These builds are working as expected in Browsers & Font Goggles.

Unfortunately it has however introduced a new issue. Within InDesign CC 2020 (15.1), these new VF binaries don't work as intended.

Screenshot 2020-06-23 at 14 20 52

As soon as you change the weight axis, it reverts to the Regular weight, and shows Red (as if the font is unsupported). Using the optical size at 17 or 18 also just gives the Max opsz output. All of the decimals (17.1 > 17.9) gives the Min opsz. 

Version 3.000 VF builds don't have this issue.

chrissimpkins:
Did the builds with bracket layers defined at 17.5 work as expected in Adobe products?

Colophon-Foundry:
I don't think they ever built as they failed the build process. I can build locally and test though?

chrissimpkins:
I don't know that we pushed them to a build branch.  They still build with fontbakery fails.  The builds have to take place for fontbakery to test them.  Mind trying a Glyphs export of the current sources with brackets defined at 18 to see if that works in InDesign?  I checked the STAT tables with these builds on Samsa and it looks like all of our named instances are properly defined and interpolation works there too.  Unclear why they are not working in InDesign...

chrissimpkins:
Edd and I met to discuss this.  The bracket layers will be moved to 18 and the brace layers will be maintained (in contrast to removal that happened with the changes in this PR) and moved to 17.  I'll open a new PR with these changes and build new files so that we can have a look.

chrissimpkins:
Error: https://github.com/googlefonts/googlesans/runs/804093082?check_suite_focus=true#step:7:133
Sources (**note**: these are not in the branch that Edd used for this PR): https://github.com/googlefonts/googlesans/tree/Latin-opsz/source/GoogleSans

@anthrotype Cosimo, it looks like we are confusing fontmake with this opsz axis transition.  The goal here is to move away from interpolation on the optical size axis.  The previous opsz range was 14 min - 18 max.  We moved the 14 masters up to 17.  As a result we needed to move the brace (defined at sizes 15,16 in various places) and bracket (defined at 16 in various places) layers to 17 and 18, respectively.  The fvar default value is defined at our regular weight max optical size master (opsz=18, weight=400).  

With these changes, it looks like we error out on the avar table generation step.  The data begin at [this line in the build output](https://github.com/googlefonts/googlesans/runs/804093082?check_suite_focus=true#step:7:133).  In the avar normalized mapping, it looks like we get something like this:

```
INFO:fontTools.varLib:Normalized master locations:
[{'Optical size': -1.0, 'Weight': 0.0},
 {'Optical size': -1.0, 'Weight': 0.0},
 {'Optical size': -1.0, 'Weight': 1.0},
 {'Optical size': -1.0, 'Weight': 1.0},
 {'Optical size': 0.0, 'Weight': 0.0},
 {'Optical size': 0.0, 'Weight': 1.0}]
```

Any thoughts about how we can address this issue?

anthrotype:
please run fontmake with `--verbose DEBUG` so we can see the exception stacktrace. 
It looks like you end up with more than one master at the same location as another, which is impossible in the OTVar model. If you end up like that, you have the opportunity to merge those duplicate masters into one; if you can't because they are different designs, then they need to have different location..

anthrotype:
the way fontmake compiles what Glyphs.app calls "brace" layers is by compiling interpolatable TTF masters that only contain those brace layer glyphs and have as their location in the designspace the coordinates defined in the brace layer definition. Then, while computing deltas for gvar, for each glyph, if it is contained in one of the masters (including both the _full_ masters and the "brace" subsets) the neutral master is subtracted from it and you obtain your delta; if it's not in a master (as "brace" masters only contain some not all glyphs) then that master is skipped.
By definition each masters -- including brace layers -- must have distinct locations. So I suggest that you use something like 17.5, or 17.99 for your brace layers, instead of 17 or 18 which are locations of the other full masters.
If you want an abrubt transition between different design then maybe instead of brace layers you want to use 'bracket' layers, which will replace one glyph with another (using GSUB feature variations) when the selected instance location falls within some axis regions.

chrissimpkins:
Thanks Cosimo.  Edd wanted to keep the brace layers.  @Colophon-Foundry will you weigh in on what we should do with the location of your brace layers here? Maybe 17.1?

Colophon-Foundry:
Brace layers would be best at 17.1
Bracket layers at 17.5 if we can!

chrissimpkins:
Thanks Edd.  I will move the brace layers to 17.1.  I don't believe that we can do the bracket layers at 17.5

chrissimpkins:
> ValueError: invalid literal for int() with base 10: '17.01'

@anthrotype glyphsLib doesn't like the non-integer brace layer values: https://github.com/googlefonts/googlesans/runs/805147793?check_suite_focus=true#step:7:129

I used 17.01 instead of 17.1 as I mentioned in the post above

chrissimpkins:
@Colophon-Foundry Edd, I took a look at the glyphsLib source and it appears that the non-integer brace layer values are not supported.  They must be integer values and, at the moment, these changes raise an exception with builds of braces anywhere between 17 and 18.  Thoughts about transitioning these to bracket layers?  Braces cannot be defined at the same location as masters.

anthrotype:
Does Glyphs app support floats in brace/bracket layers names? If do, we should support that in glyphsLib

chrissimpkins:
> Does Glyphs app support floats in brace/bracket layers names? If do, we should support that in glyphsLib

Yes.  The above comments were based on my source edits so that the brace layers were defined with the names `{17.01}`.  This was performed in the Glyphs editor

chrissimpkins:
Before you go to significant effort though, IMO we are nitpicking an issue that will almost never be viewed in this typeface.  There will be explicit guidance for the type scale that teams should use and save for situations where there is an attempt to animate the opsz axis across the 17 - 18 range, or someone sets a text size to a value between 17 and 18 (who does that?) this will simply not be seen.

chrissimpkins:
The brace layers are being used for the following:

- numeral 1 and related forms (embedded in circles, small caps, etc)
- percent and related forms (perthousand, etc)
- ampersand
- a handful of marks

@Colophon-Foundry will you please comment on whether you are comfortable bagging the brace layers altogether or converting them to brackets? We are stuck until I know how you want to move forward here.

anthrotype:
You could still open an issue on glyphsLib repo about supporting float coordinates in brace/bracket layers

chrissimpkins:
> You could still open an issue on glyphsLib repo about supporting float coordinates in brace/bracket layers

I am trying to understand if this is even a good practice.  Thoughts?  At the very least, I intended to submit a PR that improves the exception handling message to indicate that only integer values are accepted.  "invalid literal for int() with base 10: '17.01'" is somewhat cryptic.

chrissimpkins:
I removed the braces as of 26b82724f59bd4316c6f6bfa7aed667762096989 and here is where we are with the interpolation across values between 17 and 18:

GS-opsz-no-braces

I think that this is acceptable for now.  Edd, we can revisit this issue if you want to modify the design with intermediate masters in a way that the fontmake pipeline supports. I'm going to keep going with these builds and we will work changes in when they are available.

chrissimpkins:
Closing in favor of https://github.com/googlefonts/googlesans/pull/62 that supersedes this PR

davelab6:
Sounds good, thank you Chris!

chrissimpkins:
> You could still open an issue on glyphsLib repo about supporting float coordinates in brace/bracket layers

@anthrotype https://github.com/googlefonts/glyphsLib/pull/616 is the improved diagnostics for brace layer names

PR #61, opened 2020-06-22
Move bracket layer definitions to opsz 18

chrissimpkins:
PR'ing this against Edd's Latin branch so that he can view the diff between the source files.

Related #60


chrissimpkins:
Edd feels that we need to add the brace layers back based on the discussion that we had during a meeting today.  Closing this in favor of a different PR that will revert back to the master branch sources that include brace layers prior to the changes made in #60

PR #67, opened 2020-07-06
Add Thai looped and loopless designs

cadsondemakoffice:


madig:
I'm working on an import script for transferring the script-specific fonts into the main one.

* Which variant is going to make it into the big unified font, the looped or loopless one? 
* What's the state of the features? I see the looped uprights rely on an external feature file not included in this PR.
* I see `{15}` brace layers for numbers (zero, ...). Are these somehow Khmer-specific or leftovers from the time the fonts were split off the LatGrkCyr main ones?

chrissimpkins:
The looped vs loopless decision is under discussion.  The design team presents their final deliverable in a week and then it will be in the G team's hands to decide which style is default in the merged typeface.  The other style will be an independent family with separate build artifacts.  I will have more information and that decision by early next quarter.

PR #65, opened 2020-07-02
Static build name fixes following 2-axis VF name updates

chrissimpkins:
This PR fixes the static font family name regression that occurred with the new naming approach that was required to support appropriate fvar table definitions in the 2-axis variable fonts. Prior to these changes, both optical sizes show under the same family name (Google Sans) rather than the separate opsz families: Google Sans and Google Sans Text.

Changes include:

- remove WWS family and subfamily name custom parameters in the sources
- add preferredFamily and preferredSubfamily custom parameters in the sources
- add stylemapFamilyName custom parameter = "Google Sans Text Medium" to Text Medium and Text Medium Italic instances 
- update the source/Makefile with text opsz static font file path edits as the file paths regressed relative to our v3.000 release with the name table changes that occurred here

These changes are based on the `Latin-opsz` branch and work in #62

PR #70, opened 2020-07-08
Fix partial instance VF naming

chrissimpkins:
These changes fix the name tables in our partial instance (wght axis only) VF builds.  Prior to these changes there were two problems:

- the fvar table was not properly generated because the partial instancer was based on our old min opsz master definition of 14.  The instance definitions were not identified with this definition at compile time.  This value was updated to 17
- nameID 268, 269, 270, and 271 records included "Text" in the record string and this led to inappropriate style names in the min opsz family (Google Sans Text). These *should* be the values in the 2-axis VF builds, but not in the 1-axis builds. 
 These edits occur after the compile step in the build pipeline

These changes are based on the source in #65

### DIFF
--- a/scripts/gs-partial-instancer.py
+++ b/scripts/gs-partial-instancer.py
@@ -19,7 +19,7 @@ import os
 from fontTools.ttLib import TTFont
 from fontTools.varLib import instancer
 
-MIN_OPSZ_SIZE = 14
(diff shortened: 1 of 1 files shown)


chrissimpkins:
Merging into our naming update branch

Issue #79, opened 2020-09-03
Tabular zero to zero spacing is too narrow


chrissimpkins:
Reporting team: Nest Energy

In time strings, the tabular figure zero forms are too narrowly spaced when set next to each other at the top of the hour (e.g., 6:00pm).  

Mockups that demo this issue:

![Screen Shot 2020-08-17 at 1 30 43 PM](https://user-images.githubusercontent.com/4249591/92140365-fbb9c580-edde-11ea-976d-fb5e90856278.png)

chrissimpkins:
Edd fixes in https://github.com/googlefonts/googlesans/commit/f7dda56fca86b763202dc2a992a159deb4a822b6 and https://github.com/googlefonts/googlesans/commit/1e1779a19255886ade81291898cc81ccc14fda1c

chrissimpkins:
Build files distributed to product team for review.  Planning to release these updates in v3.002

chrissimpkins:
Edd's design updates merged for v3.002 release @ https://github.com/googlefonts/googlesans/pull/76/commits/d23dd1c02721e3b1e2917fa45a884819036c53d8 and https://github.com/googlefonts/googlesans/pull/76/commits/bb211835fec4656280a72cb348d8153922a5bf11

chrissimpkins:
The product team signed off on the changes.  They are very happy with these updates.  Thanks @Colophon-Foundry!

chrissimpkins:
Released in v3.002

PR #76, opened 2020-08-13
v3.002: Hebrew script merge, tabular zero spacing fix, tab figure calt feature fix

chrissimpkins:
This PR supersedes the Hebrew script merge work that we performed in #56.  The merge was repeated on the HEAD commit in the master branch as the upstream source diverged with work that was performed to wrap up the v3.001 release.

This PR merges the Hebrew script design work that was performed by Daniel Grumer and Yanek Iontef into the upright and italics source files.

In addition to the merge of the Hebrew script designs, the following changes were made based on discussions in meetings and in the #56 thread:

- edited the Latin ccmp `lookup ccmp_latn_1` lookup to **remove** the `latn` script registration
- added tabular and caps forms of the sheqel currency glyph + feature code support

During the first round merge with the Hebrew designs, Daniel recognized an environment script/language dependent issue with the glyph composition/decomposition (ccmp) feature support in the Hebrew. This issue was addressed by removing the Hebrew script + lang registration in the ccmp feature code **and** the Latin registration of a ccmp lookup that was positioned at a higher priority level in the feature source.  Edd Harrington and Gor Jihanian confirmed that this was not necessary for Latin script support.  These were merged as commented out fields from Daniel's source work and I edited them out of the ccmp feature source altogether.

Closes #47

cc @danielgrumer 
cc @Colophon-Foundry 
cc @gorjious 

---

### TODO

- [x] Fontef design team review and signoff (Yanek)
- [x] Fontef design team review and signoff (Daniel)

PR #80, opened 2020-09-03
calt fea update: fix colon position between tabular figures

chrissimpkins:
The `calt` feature code required an update to center the colon between tabular figures.  When `tnum` is on, the `colon.tf` is defined rather than `colon`.  This change updates the feature code to the substitution target `colon.tf'` from `colon'` when positioned between *.tf figures in the `calt` feature source.

This change will be merged to #76 and released in v3.002.

cc @Colophon-Foundry (just a FYI)

Related: https://github.com/googlefonts/googlesans/issues/79


chrissimpkins:
Fix confirmed in Font Goggles as of 493ddc85ad8d117c662b2b94e5e92d3427feaa42

PR #90, opened 2020-09-14
Add automated GitHub release creation and build artifact push workflow

chrissimpkins:
This workflow is executed when a `v*` tag is pushed to the repository.  It automates creation of a GitHub release with the version tag in the release name and `build` directory as an attached zip archive artifact.


chrissimpkins:
Merging to test this with the v3.001 release

PR #105, opened 2020-11-24
Glyphs.app to UFO source conversion

madig:
- Migrates Glyphs.app sources to UFO and Designspace sources.
- Keeps build system the same. 
- Adds tooling to import external (Glyphs.app) sources from vendors and keep conventions.
- Tooling automates
    - normalizing files 
        - removing work artifacts like background and other backup layers
        - removing Glyphs.app metadata, e.g. layer order
        - but keeping semantically useful data, e.g. metrics keys
        - removing non-existing or empty kerning pairs or groups
        - maintaining ufo2ft filters
    - generating correct GDEF
- New workflows:
    - Vendors that place Glyphs.app files into the sources directory and push a branch will have those files compiled into a downloadable variable fonts automatically, using the same
scripts as the main files.
    - Every commit will build downloadable Glyphs. app files from the UFO and Designspace sources. Vendors should be able to easily grab those changes.
- Accidentally fixes an anchor propagation bug in the Display statics in the previous build pipeline, keeps bug for VFs.

PR has been cleaned up, comments below refer to previous states.


madig:
So, I'm comparing TTX dumps. E.g. the Text Medium TTF has no relevant diffs.

However, both static Regular and Italic show a larger `BaseCoverage` table plus a bunch of additional `BaseRecord`s. Maybe that's caused by a different GDEF table?

Also, the variable upright font has more BRACKET glyphs in its GDEF `GlyphClassDef` and a larger `BaseCoverage` table plus a bunch of additional `BaseRecord`s. The Italic VF has a whole lot additional diffs, some GDEF `VarStore` `VarData` changes and plenty of `XAdvDevice.EndSize` changes that I can't place. Not sure what's going on there.

belluzj:
We'll integrate fixing these diffs into our planning

madig:
Found out why the Italic .glyphs and DS compilations were different. Fontmaking the Glyphs file does not propagate anchors for a bunch of BRACKET glyphs and that's where all the diffs came from.

madig:
Wait, now the statics have the same problem. The DS instances have additional GDEF entries and associated anchor data compared to the original produce 🤔 It's in the same vicinity as the offending VF glyphs, too: Gbreve, Gdotaccent, yacute, ... -- maybe something to do with `<rules>`? 🤔

PS: I noticed the `LigatureCaretByPos` in static instances are now taken from the default source in all interpolated instances. I don't think that is supposed to happen. VFs should be fine because they are compiled individually and then merged. The original pipeline has the same problem if I'm not mistaken.

Related: https://github.com/googlefonts/fontmake/issues/711

PPS: A cheap and easy way to get this done correctly is to generate VFs primarily and use Cosimo's instancer to cut static instances.

moyogo:
> Found out why the Italic .glyphs and DS compilations were different. Fontmaking the Glyphs file does not propagate anchors for a bunch of BRACKET glyphs and that's where all the diffs came from.

Regarding the anchor propagation, we do want to propagate the anchors to the bracket layer glyphs, they should behave the same in all instances, at least in the full font.

madig:
So then the DS compilation process is actually doing the right thing and the glyphsLib pipeline is bugged.

madig:
ufo2glyphs2ufo works now. There are a few things that aren't clean, e.g. some bracket glyphs get kicked out of `public.postscriptNames`, but nothing major.

scripts/gs-glyphs-norm.py
```diff
@@ -27,6 +27,8 @@ def main(argv):
 
         try:
             font = GSFont(fontpath)
+            font.customParameters["Disable Last Change"] = True
```
chrissimpkins, line 30: +1!

Didn't know that this was available. This is a good change.  Perhaps we should move this script out of this repository and into something public.  This could be of use to others once this is finished.

chrissimpkins: For source file size savings, we could consider removing all of the last changed XML fields as part of the cleanup too.  It won't be relevant at all after this change.

.github/workflows/py-lint.yml
```diff
@@ -1,6 +1,6 @@
 name: Python script lints
 
-on: [push, pull_request]
+on: [push]
```
chrissimpkins, line 3: I believe that this prevents the CI from being run at all on PR's from forked repos.  That is why I've kept the duplication.

scripts/gs-norm-ufo.py
```diff
@@ -0,0 +1,94 @@
+import argparse
```
chrissimpkins: Possible to make this public without the hard coded GS paths?  This might be useful elsewhere.

scripts/gs-norm-ufo.py
```diff
@@ -60,12 +59,52 @@
...
     ufo.lib["com.github.googlei18n.ufo2ft.filters"] = [
         {"name": "propagateAnchors", "pre": True},
-        {"name": "flattenComponents", "pre": True},
+        # Uncomment after attaining parity between Glyphs file and DS compilation.
+        # {"name": "flattenComponents", "pre": True},
     ]
+    if "Italic" in ufo_path.stem:
+        # Compiling the Italic Glyphs file does not propagate anchors in the following
+        # glyphs:
+        ufo.lib["com.github.googlei18n.ufo2ft.filters"][0]["exclude"] = [
```
chrissimpkins: Just getting to this commit.  I see, there are project-specific hard coded values in here.  Np, we'll keep it as a project script and can consider a more general version down the road.

madig: Mind you that the script may try to replicate existing bugs here, see my email.

.github/workflows/py-lint.yml
madig: Hm. All examples on https://docs.github.com/en/free-pro-team@latest/actions/guides/building-and-testing-python use a singular `push`.

scripts/gs-glyphs-norm.py
madig: Already done by the UFO normalization script :)

madig:
So, import of Lao and Thai with my script works, Khamer not so much yet. Features are not copied because they might need manual handling until we find something that works automatically. The UFO normalization script scrubs the UFOs to only leave what is relevant for building and maintenance.

The other thing I have left out is brace layer handling. E.g. the current Khmer sources have different brace layers than the base sources which are probably leftovers from the time font dev was started. Designspace `<rules>` merging has to be done manually for now I suppose...

madig:
@googlebot I consent.

madig:
@chrissimpkins please consent @ googlebot, otherwise it will keep spamming us :D

chrissimpkins:
@googlebot I consent

chrissimpkins:
It is obnoxious and the CLA doesn't even apply to this project.  I've signed the CLA in the past for other projects and I don't know why it would be flagging my commits.

chrissimpkins:
And there is no way to deactivate it.  I've tried...

.github/workflows/py-lint.yml
chrissimpkins: I believe that this happened on ufolint when you pushed a PR from a forked repo. The CI was not run. Perhaps I am wrong.  I can look into it.  Not a problem unless we accept CI ready source as PR from branches outside of this remote.  OK to remove it for now.

chrissimpkins:
> So, import of Lao and Thai with my script works, Khamer not so much yet. Features are not copied because they might need manual handling until we find something that works automatically. The UFO normalization script scrubs the UFOs to only leave what is relevant for building and maintenance.
> 
> The other thing I have left out is brace layer handling. E.g. the current Khmer sources have different brace layers than the base sources which are probably leftovers from the time font dev was started. Designspace `<rules>` merging has to be done manually for now I suppose...

Note that those are not final merge ready design sources. They are only intended for exploration in this workflow right now.  The team will deliver final Thai looped + looples, Khmer loopless, and Lao loopless in next week.   I'm not sure that I'd spend a huge amount of time troubleshooting issues until we have final design files in hand.

chrissimpkins:
Khmer and Lao looped should be available during Q1.  Awaits a budget decision on the Google side.

anthrotype:
I don't think deleting the bot comments helps to stop it, it'll just continue to spam. It was complaining that there were multiple authors (all of which individually had signed their CLA) but simply requested that all consented to having their name in the co-authored PR.
The bot is activated at an organization level on all the repositories belonging to it. I don't think we can disable it on a repository's basis. Nor whether we should try /cc @davelab6

chrissimpkins:
Thanks Cosimo.  I spoke with Dave about it and he tells me that it is not possible to deactivate it on the Google Sans repos.  The CLA does not apply to this project and these are unnecessary checks/notices.

chrissimpkins:
And that is correct.  Until the author adds the consent, deletion of the messages just prompts the bot to post another one.  You can delete them after the consent acknowledgment is available in the thread.

docs/MAINTAINER.md
```diff
@@ -46,4 +46,89 @@ $ make list-deps
...
+
+```
+$ python scripts/update-gdef.py
+```
+
+### Conventions for External Vendors
+
+To make merges into the main source base as seamless as possible, vendors should
+
+* Only keep glyphs relevant to their script in their source files and carry as few base source glyphs as possible (think numbers, punctuation also present in the base sources).
```
belluzj: I don't think that's a good recommendation; vendors should be able to see all the scripts together while they're working. Also, merging only what they want to deliver and not the rest is part of our merging job, not the vendor's responsibility.

docs/MAINTAINER.md
```diff
@@ -46,4 +46,89 @@ $ make list-deps
...
+* Clean out all miscellaneous data like kerning and groups when starting out so that only script-relevant data is present.
```
belluzj: Same, cherry-picking what to merge is our job; vendors should be able to evaluate kerning in context of other kerning for other scripts.

docs/MAINTAINER.md
```diff
@@ -46,4 +46,89 @@ $ make list-deps
...
+![External Vendor Workflow](assets/workflow.png)
+
+### Importing a Glyphs.app File
+
+First, run the import script for the upright and italic source to import.
+
+```
+$ python scripts/gs-import-glyphs-file.py --source GoogleSansSomeScript.glyphs --target source/GoogleSans/GoogleSans.designspace
+
+$ python scripts/gs-import-glyphs-file.py --source GoogleSansSomeScript-Italic.glyphs --target source/GoogleSans/GoogleSans-Italic.designspace
```
belluzj: These two scripts should take as argument a list of glyphs that vendors were meant to deliver, so that the scripts know to import only those glyphs

docs/MAINTAINER.md
```diff
@@ -46,4 +46,89 @@ $ make list-deps
...
+* Name all glyphs according to the naming standard used by Glyphs.app and ideally not change them after the first time they've been imported to the base sources (the import script remembers names to know what to always import).
+* The sources must be buildable with `fontmake`. Various advanced Glyphs.app features are off-limits because the open-source pipeline does not support them, among them smart components.
+* Not change glyphs, groups and kerning that exist in the base sources, as they will be ignored by default. If vendors want to change something that already exists, they must explicitly flag it with us.
+* Bundle up test documents for checking the correct shaping of text and application of features, to have tests for functionality after merging.
+
+### Workflow
+
+Vendors can use Glyphs.app or work directly on UFOs and Designspaces. glyphsLib is used to generate Glyphs.app files for those who need it. Vendors can go off and work on their script and come back once it is ready to be merged or commit their changes regularly to a branch, where we will do the merge process.
+
+![External Vendor Workflow](assets/workflow.png)
```
belluzj: I'll send you the latest version of that diagram (the one that was in the slides) and you can also include the two others about the merge process and the QA process in the next sections of your documentation

madig:
@chrissimpkins some vendors use their own glyph naming standards, but in my doc I tell them to use the Glyphs standard -- what should we actually do?

madig:
Need to break up scripts into gs-import, gs-export, normalize and merge (UFO to UFO).

belluzj:
> Need to break up scripts into gs-import, gs-export, normalize and merge (UFO to UFO).

can you name them gs-ufo2glyphs and gs-glyphs2ufo for clarity?

I'm going to post a diagram soon to explain what each script does.

chrissimpkins:
> @chrissimpkins some vendors use their own glyph naming standards, but in my doc I tell them to use the Glyphs standard -- what should we actually do?

Is the Glyphs standard/convention documented somewhere?

chrissimpkins:
> > @chrissimpkins some vendors use their own glyph naming standards, but in my doc I tell them to use the Glyphs standard -- what should we actually do?
> 
> Is the Glyphs standard/convention documented somewhere?

And is it broad enough to cover all of our scripts?

madig:
The documentation is what's inside Glyphs. There's a spin-off with some data removed at https://raw.githubusercontent.com/googlefonts/glyphsLib/master/Lib/glyphsLib/data/GlyphData.xml, but we found it to sometimes not have the same data as the "official" source. It also lags behind. It still covers all scripts, as none of them are terribly new additions to Unicode.

One of our designers made his own Devanagari glyph names, so there might be good reasons for vendors to diverge, maybe ask them... Jany says: 

> maybe our requirement could be relaxed to "use the official one or provide their own GlyphData.xml"

chrissimpkins:
> The documentation is what's inside Glyphs. There's a spin-off with some data removed at https://raw.githubusercontent.com/googlefonts/glyphsLib/master/Lib/glyphsLib/data/GlyphData.xml, but we found it to sometimes not have the same data as the "official" source. It also lags behind. It still covers all scripts, as none of them are terribly new additions to Unicode.
> 

It would be nice to achieve an editor agnostic approach for the design teams.  Having said that, all teams associated with the scripts that will be merged here are using Glyphs app in glyphs source format at the moment so this should not be an issue.  We should address why designers moved away from GlyphsData.xml defined names.  Is this a hypothetical at this stage or are you seeing this in sources in open PR's here?

> One of our designers made his own Devanagari glyph names, so there might be good reasons for vendors to diverge, maybe ask them... Jany says:
> 
> > maybe our requirement could be relaxed to "use the official one or provide their own GlyphData.xml"

I believe that Kalapi and Gunnar are doing the same thing in all of the S Asian scripts.  @gunnarvilhjalmsson @kalapi can you verify this and weigh in on how you are approaching glyph naming in the S Asian work?

docs/MAINTAINER.md
```diff
@@ -102,6 +102,44 @@ To make merges into the main source base as seamless as possible, vendors should
 * Notify us if they want to change something in the base sources with their sources, as we will screen the changes out otherwise.
 * Bundle up test documents for checking the correct shaping of text and application of features, to have tests for functionality after merging.
 
+Additionally, some conventions should be followed for kerning group names and feature code:
+
+* Kerning groups should contain the name of the script they pertain to. Glyphs.app naming does this automatically by using the glyph name as a group name, which includes a script tag. Example: `boBaimai-thai`. This avoids name clashes.
+* The contents of existing groups should only be changed after Chris Simpkins gave the approval.
```
belluzj: name or contents*

madig:
> Is this a hypothetical at this stage [...]

I think it is, but it may crop up at some point, so I'm just thinking out loud.

madig:
@chrissimpkins please have a look at my additions to https://github.com/googlefonts/googlesans/blob/convert-to-ufo/docs/MAINTAINER.md

madig:
The latest commits add two new GitHub Actions workflows, one to turn the Designspace sources into Glyphs.app files for download and the other to build variable fonts from Glyphs.app files lying around in the sources folder.

Both are intended to be a convenience method of sorts for vendors. Not sure they need to always run. I think the workflows can be refined when we actually get some real world use for them.

docs/MAINTAINER.md
```diff
@@ -46,4 +46,75 @@ $ make list-deps
...
+
+* Foreground, intermediate (brace) and conditional (bracket) glyphs, no draft or background layers
+    * whose metadata (lib keys) contains only semantically relevant data like Glyphs.app's metrics keys, but not color marks.
+* Groups and kerning
+* Manually merged and arranged features
+* Manually maintained font info data
+* Automatically managed UFO lib.plist files, that contain:
+    * `public.glyphOrder` for determining the order of glyphs in the final fonts
+    * `public.postscriptNames` for determining the production glyph names in the final fonts
+    * `public.skipExportGlyphs` for listing glyph names that should not be exported to the final fonts
```
chrissimpkins, line 65: Please note any that you come across.  If they are not required we can consider pruning rather than maintaining as non-exported glyphs in the set.

docs/MAINTAINER.md
```diff
@@ -46,4 +46,75 @@ $ make list-deps
...
+    * `com.schriftgestaltung.customParameter.GSFont.disablesLastChange` for telling Glyphs.app to not put "last changed" markers into glyphs, which we don't need in the UFO format
+    * `com.schriftgestaltung.fontMasterID` for making it easier to match vendor Glyphs.app files to be imported to the target UFOs
+* Manually managed Designspace `<rules>` for describing conditional (bracket) glyphs
+* Automatically managed Designspace instance and global libs, that contain:
+    * Instances:
+        * `com.schriftgestaltung.customParameters` for carrying build-relevant metadata like PANOSE values
+    * Global:
+        * `GSDimensionPlugin.Dimensions` for storing Glyphs.app's metadata for stem thicknesses
+        * `com.github.googlei18n.ufo2ft.featureWriters` for build-relevant options on how to generate OpenType layout data
+        * `public.skipExportGlyphs` for listing glyph names that should not be exported to the final fonts
```
chrissimpkins, line 80: This is a duplicate of line 65 above and this line can be deleted.

docs/MAINTAINER.md
chrissimpkins: This was resolved but the line does not appear to have changed.  How problematic is this for you? We can ask the teams to clean the sources before they push them to us for a merge and/or document what should be merged in what is pushed.

docs/MAINTAINER.md
```diff
@@ -46,4 +46,75 @@ $ make list-deps
...
+
+### Conventions for External Vendors
+
+To make merges into the main source base as seamless as possible, vendors should
+
+* Only keep glyphs relevant to their script in their source files and carry as few base source glyphs as possible (think numbers, punctuation also present in the base sources).
+* Clean out all miscellaneous data like kerning and groups when starting out so that only script-relevant data is present.
+* Ideally make no use of intermediate (brace) layers if the design allows it
+* Name all glyphs according to the naming standard used by Glyphs.app and ideally not change them after the first time they've been imported to the base sources (the import script remembers names to know what to always import).
```
chrissimpkins: What is the standard in editors that are not Glyphs app?  Do they have different naming conventions or do they follow Glyphs app lead?

docs/MAINTAINER.md
```diff
@@ -46,4 +46,75 @@ $ make list-deps
...
+* The sources must be buildable with `fontmake`. Various advanced Glyphs.app features are off-limits because the open-source pipeline does not support them, among them smart components.
```
chrissimpkins: This is a valid statement and is included in every development team's contract.  Not a controversial statement and they should abide by this (should be already)

docs/MAINTAINER.md
```diff
@@ -46,4 +46,75 @@ $ make list-deps
...
+* Not change glyphs, groups and kerning that exist in the base sources, as they will be ignored by default. If vendors want to change something that already exists, they must explicitly flag it with us.
```
chrissimpkins: Let's discuss how we want them to document this and include it in this line.  Perhaps we can create a "XXX script merge documentation" document that is delivered as part of the final design deliverable and serves as communication about this issue.  In the past, I met F2F with the designer and went through the source files with them.  They would be open to that if you'd like to discuss changes directly.  That is likely more efficient, less effort on their side but does not create documentation for future reference.

docs/MAINTAINER.md
```diff
@@ -46,4 +46,75 @@ $ make list-deps
...
+* If the vendor creates test documents for checking the correct shaping of text and application of features, we'd love to get them, to test for functionality after merging.
+
+### Workflow
+
+Vendors can use Glyphs.app or work directly on UFOs and Designspaces. glyphsLib is used to generate Glyphs.app files for those who need it. Vendors can go off and work on their script and come back once it is ready to be merged or commit their changes regularly to a branch, where we will do the merge process.
```
chrissimpkins: > Vendors can use Glyphs.app or work directly on UFOs and Designspaces

Is the following a valid change?

"Vendors can use Glyphs.app to work on *.glyphs source files or work directly on UFOs and Designspaces with any editor".

docs/MAINTAINER.md
```diff
@@ -46,4 +46,75 @@ $ make list-deps
...
+
+![External Vendor Workflow](assets/workflow.png)
+
+### Importing a Glyphs.app File
+
+First, run the import script for the upright and italic source to import.
+
+```
+$ python scripts/gs-import-glyphs-file.py --source GoogleSansSomeScript.glyphs --target source/GoogleSans/GoogleSans.designspace
```
chrissimpkins: Our macOS friends will likely need the python executable to be `python3` here...

docs/MAINTAINER.md
```diff
@@ -46,4 +46,75 @@ $ make list-deps
...
+
+$ python scripts/gs-import-glyphs-file.py --source GoogleSansSomeScript-Italic.glyphs --target source/GoogleSans/GoogleSans-Italic.designspace
```
chrissimpkins: and `python3` here

docs/MAINTAINER.md
```diff
@@ -46,4 +46,75 @@ $ make list-deps
...
+
+Next, extract the features and manually massage them into the existing sources. The font info may also need to be updated, chiefly Unicode and codepage ranges. The data can be hand-picked from a `glyphs2ufo` run on the Glyphs.app files.
+
+```
+$ glyphs2ufo GoogleSansSomeScript.glyphs -m source/GoogleSans/staging
+
+$ glyphs2ufo GoogleSansSomeScript-Italic.glyphs -m source/GoogleSans/staging
+```
+
+Important: the import script will specifically import only glyphs, groups and kerning that do not already exist in the font that is being imported into, except when the same names have been imported before. This is done to avoid changes to things that should not have been changed but have been. Usually, those changes are unintented. Vendors who want to change something existing on purpose must notify us of this ("list of changes out of scope").
```
chrissimpkins: sp: "unintented" -> "unintended"

docs/MAINTAINER.md
madig: There already are some, e.g. that Google G.ss06 thing that occupies its own layer but is not exported for some reason. See https://github.com/googlefonts/googlesans/blob/874af85/source/GoogleSans/GoogleSans.designspace#L454-L463

docs/MAINTAINER.md
madig: Glyph naming is free-form, so everyone does whatever the hell they want.

docs/MAINTAINER.md
madig: I'm just seeing that you are commenting on outdated text! :O Use https://github.com/googlefonts/googlesans/blob/convert-to-ufo/docs/MAINTAINER.md

docs/MAINTAINER.md
```diff
@@ -46,4 +46,169 @@ $ make list-deps
...
+* Keep the sources buildable with `fontmake`. Various advanced Glyphs.app features are off-limits because the open-source pipeline does not support them, among them smart components.
+* Provide a list of glyph names and group names to import into the base sources ([see below](#getting-a-list-of-glyphs-and-kerning-groups)).
+* Notify us if they want to change something in the base sources with their sources, as we will screen the changes out otherwise.
+* Bundle up test documents for checking the correct shaping of text and application of features, to have tests for functionality after merging.
+
+Additionally, some conventions should be followed for kerning group names and feature code:
+
+* Kerning groups should contain the name of the script they pertain to. Glyphs.app naming does this automatically by using the glyph name as a group name, which includes a script tag. Example: `boBaimai-thai`. This avoids name clashes.
+* The name or contents of existing groups should only be changed after Chris Simpkins gave the approval.
```
chrissimpkins: "Chris Simpkins gave the approval" -> "Google approval".  Who knows how long Dave intends to keep me around :)  This will be an evergreen statement that requires no maintenance down the road.

docs/MAINTAINER.md
```diff
@@ -46,4 +46,169 @@ $ make list-deps
...
+* Feature code should be organized so that:
+    * feature blocks and lookups are declared separately. This makes merging much easier.
+    * lookups should have descriptive names and should include, where it makes sense, the language and feature tag where they are used. Example: `NLD_locl_ij_substitution` for a netherlandish lookup that replaces `i' j'` by `ij`. See https://docs.microsoft.com/en-us/typography/opentype/spec/languagetags for ISO 639 language tags.
+
+Feature file example:
```
chrissimpkins, line 143: 👍 Love the example of how it should be done.

docs/MAINTAINER.md
```diff
@@ -46,4 +46,169 @@ $ make list-deps
...
+
+Vendors can use Glyphs.app or work directly on UFOs and Designspaces. glyphsLib is used to generate Glyphs.app files for those who need it. Vendors can go off and work on their script and come back once it is ready to be merged or commit their changes regularly to a branch, where we will do the merge process.
+
+![Script Workflow](assets/scripts.png)
+
+### Importing a Glyphs.app File
+
+![External Vendor Glyphs Workflow](assets/new_situation_glyphs.png)
+
+First, run the `gs-glyphs2ufo.py` script for the upright and italic source to convert the source files and place them into `source/GoogleSans/staging/`.
```
chrissimpkins, line 212: Let's keep the script directory path on all of these new scripts so that they are easy to locate.  Or a one-liner somewhere in the documentation for designers that says "scripts that you will need live in xxx"

docs/MAINTAINER.md
```diff
@@ -46,4 +46,169 @@ $ make list-deps
...
+
+```
+$ python scripts/gs-glyphs2ufo.py GoogleSansSomeScript.glyphs --target-dir source/GoogleSans/staging/
```
chrissimpkins: defaults to Py2.7 on macOS

docs/MAINTAINER.md
```diff
@@ -46,4 +46,169 @@ $ make list-deps
...
+
+$ python scripts/gs-glyphs2ufo.py GoogleSansSomeScript-Italic.glyphs --target-dir source/GoogleSans/staging/
```
chrissimpkins: defaults to Py2.7 on macOS

docs/MAINTAINER.md
```diff
@@ -46,4 +46,169 @@ $ make list-deps
...
+```
+
+Next, import the resulting Designspaces into their intended target Designspaces:
+
+```
+$ python scripts/gs-merge-designspace.py \
```
chrissimpkins: more python -> python3 interpreter areas in this block of docs.  Even if your scripts support Py2.7, it may be worth switching to the python3 executable convention because anything based on fonttools will not work with `python` on macOS.

docs/MAINTAINER.md
chrissimpkins: Lol.  OK, then this is at least a convention. To support a designer who is working in RoboFont, do we have an approach to find glyph names that is not "well, just open your Glyphs editor and figure it out"?

docs/MAINTAINER.md
chrissimpkins: I was trying to comment inline and it was a bit difficult to find in the 5000+ changed files... Found the GH filter.

docs/MAINTAINER.md
madig: I think we should prune what we can :) when do we start? :D

docs/MAINTAINER.md
madig: This same key is stored in UFO and Designspace due to the fragile link between the two formats. They must match.

docs/MAINTAINER.md
madig: That's actually going to be the simplest solution. The other is to peruse the XML dump in https://github.com/schriftgestalt/GlyphsInfo for the right name.

docs/MAINTAINER.md
madig: Hm... maybe we could make a GitHub PR template they just fill in? That way we don't have to make extra documents floating around and text can be updated in the PR itself?

https://docs.github.com/en/free-pro-team@latest/github/building-a-strong-community/creating-a-pull-request-template-for-your-repository

docs/MAINTAINER.md
madig: Hm. I'll change it, but since these scripts need our requirements.txt stuff, they should be run inside a venv, where both `python` and `python3` are typically available.

docs/MAINTAINER.md
chrissimpkins: Let's review the list of everything that appears to be unnecessary and discuss.

docs/MAINTAINER.md
chrissimpkins: I like it.  It wasn't clear to me that the designers would be opening the PR on the repository.

docs/MAINTAINER.md
chrissimpkins: I received an email re: this being executed in a venv but I can't locate it in this thread at the moment. Of course you are correct about this.  

How about a different suggestion then so that they do not need to activate/deactivate the venv and we provide a check that they are actually using the venv if they follow the instructions (i.e., it won't work unless the venv has been built at some stage in the past, may not be up to date but this may be the best that we can do...):

```
.venv/bin/python scripts/gs-merge-designspace.py
```

madig:
I'm digging into the static font diffs and found: the problem is actually not in the GDEF, which is the same, but in the anchor propagation. All Display (non-Text) statics lack anchors in e.g. `Gbreve` in the old pipeline but have them in the new one. There is no diff between old and new Text statics and between old and new VFs. From a quick glance, glyphs which are replaced by Designspace rules (Text -> Display) are affected.

This leads me to think that this is probably caused by Designspace rules swapping glyphs interfering with anchor propagation in the old pipeline, landing them on the existing (accidental) blacklist when they shouldn't.

This means the old pipeline would have two problems:

1. Anchors aren't propagated to conditional (bracket) glyphs when they should
2. Designspace rules swap glyphs for the statics, landing them on the (accidental) list of ignored glyphs of point 1.

The new pipeline solves both problems by doing nothing.

It also gives me the idea of a FontBakery check that tries to reconstruct anchors for all glyphs for a bundle of fonts and checks if all glyphs have the same anchors.

@anthrotype

docs/MAINTAINER.md
madig: Jany told me that he was considering that we open PRs ourselves instead. I suppose that designers can also just list their intended changes in the email they'll send us or comment on the PR itself?

docs/MAINTAINER.md
madig: These are the glyphs: 

image

The `apple` glyph does not look very Apple-y. The descender glyphs are insidious: not exporting them means the glyphs that use them are decomposed, driving up file size. The non-export is probably done for aesthetic reasons so the glyphs can't show up in an Adobe glyph picker, but you can't have it both ways.

docs/MAINTAINER.md
madig: Are you fine with the rules set out for kerning group names and features? I think you mentioned that you had some ideas for kerning group names yourself.

docs/MAINTAINER.md
chrissimpkins: @alexeiva Mind addressing [Nikolaus's comment](https://github.com/googlefonts/googlesans/pull/105#discussion_r544932741) above about the _descender-cy and _descender-cy.case glyphs that are included in the set but are **not** exported?

> The descender glyphs are insidious: not exporting them means the glyphs that use them are decomposed, driving up file size. The non-export is probably done for aesthetic reasons so the glyphs can't show up in an Adobe glyph picker, but you can't have it both ways.

chrissimpkins: ping @alexeiva

docs/MAINTAINER.md
chrissimpkins: I've not defined kerning group names.  Only glyph class names.  Agree with your approach here.

docs/MAINTAINER.md
chrissimpkins: How about an issue thread for the script X merge where the vendor DnD a zip archive of the sources and associated documentation into the thread post?  You can then take these files and draft the PR for the merge.

Ideally, I think that we take email out of the process as much as possible.  Emails tend to lead to email conversation threads that are better handled/archived on repo IMO.

chrissimpkins:
> Use statmake to generate STAT table.

From the list in your OP. Is this still planned and, if so, can you explain why you prefer this to the fontTools lib approach that we are using now?  It adds a new project dependency which isn't a huge deal but I'd like to understand why it is necessary.

madig:
Oh that was a tentative idea. There's no need if scripts are in place already. I'll clean up the PR and change the description.

docs/MAINTAINER.md
Colophon-Foundry: I believe the icon glyphs are originally from product sans but were removed back before v2 as they weren’t needed. 

The descender glyphs also can be deleted but let’s double check with @alexeiva

chrissimpkins: Thanks Edd!

Re: icons, I've asked internally and will let you know when I hear back.  

Re: cyrillic glyphs, agree let's see what Alexei has to say.

madig: (I think if things are removed or set to export again, we should do so in a separate PR to keep it focused)

docs/MAINTAINER.md
belluzj: Yes, having all the files dropped in a GitHub issue sounds great, and yes we will handle the opening of a PR with the changes at the UFO level. We could write that down in the vendor documentation?

chrissimpkins: > We could write that down in the vendor documentation?

sgtm. I'll add this to vendor docs.

chrissimpkins:
I spoke with Edd Harrington last evening.  He confirmed that he has everything that he needs from the remote glyphs formatted source files.  

@madig I will go through the threads in this PR and open new issues with anything that requires further discussion.  Are you ready to merge this?

docs/MAINTAINER.md
chrissimpkins: OK to remove the G.ss06, peace, frowning face, smiling face, and apple glyphs.  Let's do this in a separate PR.

Let's keep the *-cy descenders until we hear back from Alexei about this issue.

chrissimpkins:
@madig will you please edit the new GitHub Actions workflows `build-incoming-files.yml` and `generate-glyphs-file.yml` to only upload artifacts on a push to a dedicated branch?  I don't believe that we need these artifacts on every commit across every branch and there is a cost to do so.

The config that I have in https://github.com/googlefonts/googlesans/blob/f1ca8a3719bd54b91bc59cf8391a6a126753f51a/.github/workflows/upload-artifacts.yml#L2-L8 will do this for the entire workflow if you want to use a similar approach.  I haven't come across documentation about how to do this for steps in a workflow.  We can split the artifact upload out as a separate workflow based on the branch where the commit occurs if necessary.

chrissimpkins:
@madig will you please add the Apache License v2 stub as a commented out header in all new Python scripts in this PR?  All project scripts are being released under this license.  There is no license associated with the overall project. Thanks!

docs/MAINTAINER.md
chrissimpkins: Added in initial draft of vendor docs.

docs/MAINTAINER.md
chrissimpkins: #122

chrissimpkins:
> @madig will you please add the Apache License v2 stub as a commented out header in all new Python scripts in this PR? All project scripts are being released under this license. There is no license associated with the overall project. Thanks!

https://github.com/googlefonts/googlesans/pull/105/commits/b4cec38ad2a6e9e5a0ab1724561d905184a96cd5

Thanks Niko!

chrissimpkins:
Open discussion threads in this PR will transition to issue threads:

- anchor propagation (#109)
- removal of unexported glyphs (#122)

chrissimpkins:
Testing our new GH Action workflows with a push to `build-change-to-ufo` branch.  The new Generate Glyphs.app File fails with zero exit status code when there is no glyphs source file present on the expected path.  CI will pass in the common case when there is no vendor glyphs source available during execution.  This is gated to run on master and any `build-*` branch.  I think that we can remove the master branch runs because we will not commonly have a glyphs source file on this path.  I'll take care of this in a new PR.

https://github.com/googlefonts/googlesans/issues/123

chrissimpkins:
We are ready to merge here.  A huge thanks to Nikolaus and Jany for all of the work to make this UFO source format transition and support a new merge workflow for vendor teams.  This is a great improvement in the project.

chrissimpkins:
@NeilSureshPatel Neil, this is the PR with the UFO source transition and new scripts to support the glyphs format i18n design -> UFO merges.  The master branch has our updated Makefiles/build workflow.  Let me know if you have any questions.

PR #120, opened 2021-01-15
Add vendor documentation, style guidelines

chrissimpkins:
Adds vendor docs (`docs/VENDOR.md`) on how to contribute to the project with the new UFO source format and contribution workflow added in #105

### DIFF
--- a/docs/STYLE.md
+++ b/docs/STYLE.md
@@ -48,7 +48,7 @@ This document details the source style guidelines for the Google Sans project.
 - All names should be lowercase, separated by underscores, except for the string that refers to glyph names. E.g. `Omega` should stay as is if it refers to the upper case glyph name
 - Kerning groups should contain the name of the script they pertain to. This avoids name clashes. The format is script_key_glyph_or_description
 
-#### Adobe feature code lookup names
+#### Adobe OpenType feature file lookup names
 
 - Adobe feature code lookup names should use the following format:
 
@@ -62,7 +62,7 @@ This document details the source style guidelines for the Google Sans project.
 - When a lookup is used in several features, list all feature tags in its name
 - Except if listing everything is too cumbersome and counterproductive, then drop that part of name but leave a comment instead, just above the lookup definition, to explain which scripts/languages/features are concerned
 
-### Adobe feature file source
+### Adobe OpenType feature file source
 
 - feature blocks and lookups should be declared separately
 - lookups should have descriptive names and include, where appropriate, the language and feature tag where they are used. Example: `nld_locl_ij_substitution` for a Netherlandish lookup that replaces `i' j'` by `ij`
@@ -100,7 +100,7 @@ This document details the source style guidelines for the Google Sans project.
 - A_B_C_D: `cyrl_bgr_locl_alternates`
 - C_D: `frac_precomposed`
   
-### Adobe feature file examples
+### Adobe OpenType feature file examples
 
 ```fea
 languagesystem DFLT dflt;
--- a/docs/STYLE.md
+++ b/docs/STYLE.md
@@ -12,11 +12,11 @@ This document details the source style guidelines for the Google Sans project.
 
 ### Script Tags
 
-"Script tags" in this document refer to the script tag strings defined in [OpenType Script Tags](https://docs.microsoft.com/en-us/typography/opentype/spec/scripttags).
+"Script tags" refer to the script tag strings defined in [OpenType Script Tags](https://docs.microsoft.com/en-us/typography/opentype/spec/scripttags).
 
 ### Language Tags
 
-"Language tags" in this document refer to the language tag strings defined in [OpenType Language System Tags](https://docs.microsoft.com/en-us/typography/opentype/spec/languagetags).
+"Language tags" refer to the language tag strings defined in [OpenType Language System Tags](https://docs.microsoft.com/en-us/typography/opentype/spec/languagetags).
 
 ## Style
 
--- a/docs/STYLE.md
+++ b/docs/STYLE.md
@@ -1,0 +1,131 @@
+# Google Sans Style Guidelines
+
+This document details the source style guidelines for the Google Sans project.
+
+## Contents
+
+- [Definitions](#definitions)
+- [Style](#style)
+- [Examples](#examples)
+
+## Definitions
+
+### Script Tags
+
+"Script tags" in this document refer to the script tag strings defined in [OpenType Script Tags](https://docs.microsoft.com/en-us/typography/opentype/spec/scripttags).
+
+### Language Tags
+
+"Language tags" in this document refer to the language tag strings defined in [OpenType Language System Tags](https://docs.microsoft.com/en-us/typography/opentype/spec/languagetags).
+
+## Style
+
+### Names
+
+#### Glyph names
+
+- Name all glyphs according to the naming standard used by the Glyphs.app GlyphData.xml database
+- Use underscores to separate glyph names in ligature glyph forms
+- Name localized glyph forms with the glyph name, a period, the string `locl`, followed by an all caps localized language tag suffix
+- In scripts that support upper case forms, please use title case glyph names for upper case glyph forms
+- Please do not change glyph names after they have been imported into the upstream source files
+
+#### OpenType class names
+
+- OpenType class naming should use the following format:
+  - A_B_C_D
+    - A - Context-dependent script system tag (only for Non-Latin scripts)
+    - B - Context-dependent language system tag
+    - C - Context-dependent feature tag
+    - D - Class description string
+- When a class is used by several scripts, list all scripts in the name
+- When a class is used by several languages, list all languages in the name
+- When a class is used in several features, list all feature tags in its name
+- Except if listing everything is too cumbersome and counterproductive, then drop that part of name but leave a comment instead, just above the class definition, to explain which scripts/languages/features are concerned
+
+#### Kerning group names
+
+- All names should be lowercase, separated by underscores, except for the string that refers to glyph names. E.g. `Omega` should stay as is if it refers to the upper case glyph name
+- Kerning groups should contain the name of the script they pertain to. This avoids name clashes. The format is script_key_glyph_or_description
+
+#### Adobe feature code lookup names
+
+- Adobe feature code lookup names should use the following format:
+
+- A_B_C_D
+  - A - Context-dependent Script system tag - only for Non-Latin scripts
+  - B - Context-dependent language system tag
+  - C - Feature tag
+  - D - Lookup description string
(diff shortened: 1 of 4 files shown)


chrissimpkins:
@madig @belluzj Initial draft of source style guide https://github.com/googlefonts/googlesans/blob/chrissimpkins-vendor-docs/docs/STYLE.md

docs/STYLE.md
```diff
@@ -0,0 +1,131 @@
...
+
+## Style
+
+### Names
+
+#### Glyph names
+
+- Name all glyphs according to the naming standard used by the Glyphs.app GlyphData.xml database
+- Use underscores to separate glyph names in ligature glyph forms
+- Name localized glyph forms with the glyph name, a period, the string `locl`, followed by an all caps localized language tag suffix
```
ohbendy, line 29: - In the Myanmar blocks, many of the localised forms of glyphs are encoded separately and have their own glyph names in the glyphData.xml. Also some glyphs represent different characters in different languages, and some languages can represent a letter by more than one codepoint/character/glyph (a user of Khamti language can choose from 101B, 101B FE00, AA73, AA73 FE00, AA7A or AA7A FE00 to represent letter Ra).
- Some of the languages of Burma are not accessed with the locl feature but by using variation selectors or stylistic sets (e.g when there's no OT language tag defined).
- Many localised forms of glyphs are used by several languages.
- Glyph names can already be very long without language suffixes.

docs/STYLE.md
```diff
@@ -0,0 +1,131 @@
...
+#### OpenType class names
+
+- OpenType class naming should use the following format:
+  - A_B_C_D
+    - A - Context-dependent script system tag (only for Non-Latin scripts)
+    - B - Context-dependent language system tag
+    - C - Context-dependent feature tag
+    - D - Class description string
+- When a class is used by several scripts, list all scripts in the name
+- When a class is used by several languages, list all languages in the name
```
ohbendy, line 46: In Burmese, many of the classes will be used by all languages the font supports. Currently for Google Sans we are supporting about 20 languages, so the class names would be extremely long.

belluzj: Indeed in that case it wouldn't make sense to list all of them in the class name. We have a general point two lines below in the list that says:

> - Except if listing everything is too cumbersome and counterproductive, then drop that part of name but leave a comment instead, just above the class definition, to explain which scripts/languages/features are concerned

It would apply to your example, and you could write a comment above your Burmese class definitions to explain for example that all of the 20 languages are using that class.

docs/STYLE.md
belluzj: About points 1 and 2: if there's already a convention in place in glyphsData.xml that covers the naming of all the localized glyphs that you will be delivering, then it's best to stick with that convention.

About points 3 and 4: same as in your comment about class names; if listing all the languages that are concerned is counterproductive then we could maybe leave the suffix to be ".locl" without any language specification? We could clarify that option in the document.

In general these guidelines are here to facilitate communication and promote consistency; if in a particular case they're actually hurting communication (e.g. by making glyph names too long) then it's ok to work around the guidelines. In any such case, a comment is always welcome, to explain what isn't obvious anymore from the glyph name (e.g. which languages need this glyph in your example).

Does that sound reasonable?

ohbendy: Absolutely. I was asked to comment so just wanted to flag there will be a lot of cases where the Burmese doesn't quite follow the rules. When we know what glyphs are going in the fonts, we can figure out the best naming scheme that makes sense for everyone. I created the naming scheme Glyphs uses for Burmese, so there shouldn't be too much disparity.

belluzj: @chrissimpkins I added a commit to your branch to clarify this. I hope that's OK: https://github.com/googlefonts/googlesans/pull/120/commits/7fd16b6f6b6b9f8a6d0808869414c5bff774fe71

chrissimpkins: > @chrissimpkins I added a commit to your branch to clarify this. I hope that's OK: [7fd16b6](https://github.com/googlefonts/googlesans/commit/7fd16b6f6b6b9f8a6d0808869414c5bff774fe71)

No problem at all Jany.  Thanks for adding it!

chrissimpkins: @ohbendy Ben, I agree with Jany's comments.  This is more about trying to create some semblance of consistent formatting so that we avoid chaos that will be difficult to maintain and understand down the road.  We don't want to cause problems in established downstream workflows.  I'll give the style guidance another once over and see if I have anything else to add that Jany didn't already address in his latest commit.  We'll make sure that there is plenty of flexibility in the approach but that flexibility will require copious annotation so that someone who is not us down the road can understand what we did.

chrissimpkins:
Kalapi is reviewing the documentation here in preparation for the upcoming Devanagari merge. I will be speaking with him on Thursday and will incorporate any feedback that he has before we merge this.

@gorjious Now that you've gone through the entire merge process with your Armenian script designs, do you happen to have any additional feedback on how we might improve documentation for contributors?

gorjious:
> Kalapi is reviewing the documentation here in preparation for the upcoming Devanagari merge. I will be speaking with him on Thursday and will incorporate any feedback that he has before we merge this.
> 
> @gorjious Now that you've gone through the entire merge process with your Armenian script designs, do you happen to have any additional feedback on how we might improve documentation for contributors?

I would say the workflow section should include instructions/examples of the `ufo2glyphs` script with a note that the features will be disabled in the resulting glyphs files ( unless there is a way to include/reference the features file(s) ).

I was having issues with editing and saving/exporting ufos using Glyphs then rebuilding with fontmake, which Nikolaus mentioned is known issue with Glyphs UFO export...it may be helpful to include a note that mentions this.

It would also be helpful to have a section that explains what needs to be included if something is remerged. For example, should every remerge include an accompanying `import_groups.txt` and `gs_somescript_glyph_list` of the applicable glyphs/kern groups intended for merging?

Not sure if the `gs-print-kerning-groups.py` now includes glyphs that don't have a kerning group name. During my process, I had to add the left and right kern group names manually after noticing that certain glyphs weren't importing with the script. It also feels tedious and prone to error to manually go through and delete the kerning groups which are not relevant...is it possible to have a script that the user can select all the glyphs (from the Glyphs App editor window) that they intend to merge then it will create a list of the only the kerning groups that those select glyphs have kerning pairs with? 

For the style guide, I organized my lookups using two general categories:
`lookup ..._substitution` i.e `sub @armn_uppercase by @armn_smallcaps;`
`lookup ..._contextual_substitution` i.e.` sub @armn_lowercase semicolon' by semicolon.loclARM;`
in order to differentiate between the two different types of substitutions. If this naming makes sense then it could be useful to include that.

Hope that helps.

docs/VENDOR.md
```diff
@@ -0,0 +1,42 @@
...
+## Style
+
+Please refer to [STYLE.md](STYLE.md) for source style guidelines.  The document includes information on how to format your source files in order to contribute to this project.
+
+## Requirements for Vendors
+
+We ask the following from vendors who contribute to the Google Sans project:
+
+* Follow type design conventions
+* Use a required directory and file path structure
```
madig, line 32: What is meant here?

chrissimpkins: As I understood it you wanted vendors to place their glyphs sources into a defined directory path.  We also spoke about having them simply drag and drop a zip archive with Glyphs sources into an issue thread.  What works best for you?

chrissimpkins:
Thanks @gorjious!

I will go back through these docs and see how I can work in something to address the issues that you point out.  

@madig things good on the import / export script front for the merge next week?  Gor identified a few issues that we ran into with Armenian.

madig:
Yes all good. Adding some stuff to the PR.

madig:
@googlebot I consent.

madig:
> Not sure if the gs-print-kerning-groups.py now includes glyphs that don't have a kerning group name. During my process, I had to add the left and right kern group names manually after noticing that certain glyphs weren't importing with the script. 

Hm, I'd have to look at this again. The script imports all kerning pairs with anything listed in the kerning group list and glyph list, so should cover cases like `(hyphen-arm, hyphen-arm)`, as long as that glyph is listed for import.

chrissimpkins:
> > Not sure if the gs-print-kerning-groups.py now includes glyphs that don't have a kerning group name. During my process, I had to add the left and right kern group names manually after noticing that certain glyphs weren't importing with the script. 
> 
> 
> 
> Hm, I'd have to look at this again. The script imports all kerning pairs with anything listed in the kerning group list and glyph list, so should cover cases like `(hyphen-arm, hyphen-arm)`, as long as that glyph is listed for import.

Let @kalapi and I know if you can anticipate any Devanagari information that you need now.  Sources should be heading to you on Monday.

madig:
I don't anticipate anything other than what was needed for Armn already. 🤔

madig:
> is it possible to have a script that the user can select all the glyphs (from the Glyphs App editor window) that they intend to merge then it will create a list of the only the kerning groups that those select glyphs have kerning pairs with?

On it.

madig:
One thing to note is that feature editing is "disabled" in Glyphs.app because the feature files all diverge due to an explicit GDEF table. It's possible for vendors to just load all the features and delete the GDEF stuff, then carry on.

chrissimpkins:
> One thing to note is that feature editing is "disabled" in Glyphs.app because the feature files all diverge due to an explicit GDEF table. It's possible for vendors to just load all the features and delete the GDEF stuff, then carry on.

This applies to a round trip from production UFO -> glyphs format and before a designer begins work with the glyphs files?

madig:
Before testing or implementing features. Yes, it's a mess and all because we must store an explicit GDEF.

chrissimpkins:
Better to ask that they only work in UFO's for anything based on the production sources?  This will include initial branches for the start of new work and all iterations on design work after an upstream merge is underway.  Let's decide this and then I'll get it into the documentation here.

cc @kalapi We'll go through these issues in more detail re: the upcoming Devanagari merge when we get together later today.  I'm interested to hear your feedback about the planned approach, current style guidance here.  Will this work for you?

madig:
If vendors need to test features back and forth after a merge started, working on the fea files only is probably ok. They'll have to import the features though at the beginning I guess.

madig:
Hm, this is unmerged, are you going to do something else?

chrissimpkins:
> Hm, this is unmerged, are you going to do something else?

This was largely a revision of vendor targeted docs from MAINTAINER.md into a separate file.  I'm not sure that it is necessary.

chrissimpkins:
Is there anything that you need in this PR?

madig:
I don't think so, but I see you did some small additions here and there?

PR #137, opened 2021-01-26
Experiment with HarfBuzz regression testing

madig:
---

@chrissimpkins edits below

### TODO

- [x] add direction parameter to support hb buffer direction configuration for RTL scripts?
- [x] add maintainer docs (or open issue to add in a separate PR) - added in #140 
- [x] Is uHB going to be maintained?, who is the primary maintainer?, is it the best project to base our shaping regression strategy on?


qa/shaping/armenian/ccmp.txt
```diff
@@ -0,0 +1 @@
+Է՞է՞ Ե՞ե՞ և՞
```
belluzj: I like the idea of 1 file/test case, with matching names but different extensions for the input and output. You could glob for all such files in the fontbakery test.

However I think .txt as an input format will prove limiting, as in this case you're specifying ccmp: true in the fontbakery check, it could be put in this file if it was toml or yaml.

```toml
string = "Է՞է՞ Ե՞ե՞ և՞"
comment = "Test that blabla (not mandatory but could be nice to write if what we're testing is not obvious)"
[features]
ccmp = true
```

qa/check-fea.py
```diff
@@ -322,5 +324,73 @@ def com_google_fonts_check_googlesans_features_variable_italics(ttFont):
         )
 
 
+@check(id="com.google.fonts/check/googlesans/features/armenian/min_opsz",
+    conditions=["is_not_italic", "is_not_variable_font"],
+)
+def com_google_fonts_check_googlesans_features_armenian_min_opsz(ttFont):
```
belluzj: My approach here would be to glob all the folders recursively, for all scripts (not just Armenian), but I don't know from the fontbakery perspective, is that OK to have one check that goes through a massive list of asserts? Also, you might want to have some tests run only on italic files, or only on the variable font etc. (I see that this current tests says "is_not_italic", "is_not_variable_font"). Could you put that in the TOML files of the test inputs?

Also at some point we said that maybe we could always check the variable font and the statics at the same locations to make sure that both match. Should this be automatic?

Also, given 1 input string, we might want to have several outputs depending on the location in the designspace (more weight = different x/y positions of glyphs; different opsz = different glyph names because of conditional glyphs). How to encode that into the output file? 1 big JSON such as below? Smaller JSONs for each location, with the location in the file name?

```json
[
  {
    "location": {
      "wght": 100,
      "opsz": 17
    },
    "output": [
      {
        "glyph": "uni0537",
        "cluster": 0,
        "x_offset": 0,
        "y_offset": 0,
        "x_advance": 619,
        "y_advance": 0
      },
      ...
    ]
  },
  {
    "location": {
      "wght": 400,
      "opsz": 17
    },
    "output": [...]
  },
  ...
]
```

qa/check-fea.py
```diff
@@ -322,5 +324,73 @@ def com_google_fonts_check_googlesans_features_variable_italics(ttFont):
...
+    buf.add_str(text)
+    buf.guess_segment_properties()
+    hb.shape(font, buf, features)
+
+    infos = buf.glyph_infos
+    positions = buf.glyph_positions
+
+    return [
+        {
+            "glyph": font.get_glyph_name(info.codepoint),
```
belluzj: For a second I was concerned this line wouldn't return the right glyph names because I thought "the glyphs from features won't have a code point" but it looks like from your example that it works for `uni055E.cap.comb` which I suspect doesn't have a code point?

qa/harfbuzz.py
```diff
@@ -0,0 +1,69 @@
...
+    import argparse
+    import json
+    import os
+    from pathlib import Path
+
+    parser = argparse.ArgumentParser()
+    parser.add_argument("font_path", type=Path)
+    parser.add_argument("text_path", type=Path)
+    parser.add_argument(
+        "--features", help="Comma-separated feature list like `ccmp=true,smcp=false`"
```
belluzj: The features should be in the input files

qa/harfbuzz.py
```diff
@@ -0,0 +1,69 @@
...
+
+
+if __name__ == "__main__":
+    import argparse
+    import json
+    import os
+    from pathlib import Path
+
+    parser = argparse.ArgumentParser()
+    parser.add_argument("font_path", type=Path)
```
belluzj: I also think the information about which fonts can be checked with this input file should be in the input file, as conditions such as the fontbakery stuff (is_italic, is_not_variable...)

qa/harfbuzz.py
```diff
@@ -0,0 +1,69 @@
...
+    parser.add_argument("text_path", type=Path)
```
belluzj: rename to input path. Take several entries at once so we can glob in the shell to regenerate a bunch of files at once.

qa/harfbuzz.py
```diff
@@ -0,0 +1,69 @@
+from typing import Any, Dict, List
```
belluzj: Maybe rename this script to "update_shaping_test_data" or something like that?

qa/check-fea.py
belluzj: Note: "location" is good if you test the VF but not if you test a static weight

```json
{
  "file": "GoogleSans-Regular.ttf",
},
{
  "file": "GoogleSans[opsz,wght].ttf",
  "location": {
    "opsz": 17,
    ...
```

belluzj: Note: shaping between static weight and VF can be off-by-one?

chrissimpkins: > but I don't know from the fontbakery perspective, is that OK to have one check that goes through a massive list of asserts

It is so long as there is informative output when any of the assertions fail and the test is structured so that all assertions run with the FAIL raised at the end of testing + all failed tests in the report.

Is going with pytest a better approach for the shaping regression tests?  If that is a better solution so that we have access to parameterization, fixtures, etc. or to address issues like:

> NOTE: copy-pasta from harfbuzz.py because fontbakery doesn't knwo how to import it.

 we could do that.

qa/check-fea.py
madig: HarfBuzz works with glyph ids and can reconstruct the name somehow, the glyph in question does not have a codepoint.

qa/check-fea.py
madig: Jany and me found that it's probably better to keep expected shaping results separate for statics and VFs as they differ in glyph set. The bracket glyphs are standalone in VFs but are swapped into their target glyphs in the statics. This will show up in HarfBuzz's output.

I think going with fontbakery is fine, as the error messages can contain the file name and the input that failed. I'll continue experimentation.

belluzj:
I see that you've put the input and the output in the same text file? I'm not sure that's a great idea; I would at least separate them a bit better with top-level keys `input: { text:, features: }, output: { GoogleSansRg: ... }`

(the idea being that the input should be easy to edit by hand, and the output not so much. As soon as one file will have a full sentence of input, the output section will be gigantic and editing that file will be annoying)

belluzj:
Otherwise, I think you're ready to add all the data from Gor's test page and see whether the system you've put in place is handy to use

madig:
@chrissimpkins ping. What do you think? I dub these HarfBuzz unit tests.

I'd write a json file with the `input` stuff and fill it in with `python qa/update_shaping_test_data.py myfile.json build/GoogleSans/variable/expert/*.ttf build/GoogleSans/static/expert/*.ttf`. What would need some more thinking is how to display errors. We had the idea that fontbakery could create HTML pages that sort of look like what Gor provided and mark diffs in red.

madig:
(The newly imported changes change shaping output)

madig:
@chrissimpkins ping :)

I had the idea to add some shaping options. Sometimes, you don't care about advance widths or offsets but only about the raw glyph name stream (i.e. for `rvrn` testing) 🤔

qa/check-fea.py
```diff
@@ -322,5 +328,132 @@ def com_google_fonts_check_googlesans_features_variable_italics(ttFont):
...
+            )
+        else:
+            shaped_texts = shape_static(
+                tt, shaping_texts, shaping_script, shaping_language, shaping_features
+            )
+
+        if shaped_texts == shaping_texts_expected:
+            yield PASS, "No regression detected"
+        else:
+            yield FAIL, f"Expected and actual shaping not matching for {shaping_file}."
```
chrissimpkins: Worth tossing a [difflib.unified_diff](https://docs.python.org/3/library/difflib.html#difflib.unified_diff) in here?  Then we know what the diff is in CI output without needing to manually build files, dump the shaping data and diff against the expected data.

qa/check-fea.py
```diff
@@ -322,5 +328,132 @@ def com_google_fonts_check_googlesans_features_variable_italics(ttFont):
...
+            shaped_texts = shape_variable(
+                tt, shaping_texts, shaping_script, shaping_language, shaping_features
+            )
+        else:
+            shaped_texts = shape_static(
+                tt, shaping_texts, shaping_script, shaping_language, shaping_features
+            )
+
+        if shaped_texts == shaping_texts_expected:
+            yield PASS, "No regression detected"
```
chrissimpkins: I think that this test exits your `for shaping_file in shaping_basedir.glob("*.json")` loop on the initial pass through it.  I need to keep going to understand what is in the separate shaping json files that are being iterated through.  Possibly by script area?

I think that we may need to pull this out of the loop and maintain the fails in a data structure, then yield the PASS or FAIL after all json files are tested and we know the status across all tests.

qa/check-fea.py
```diff
@@ -322,5 +328,132 @@ def com_google_fonts_check_googlesans_features_variable_italics(ttFont):
...
+    font_path: str,
+    text: str,
+    script: str,
+    language: str,
+    features: Dict[str, bool],
+    variations: Optional[Dict[str, float]] = None,
+) -> List[Dict[str, Any]]:
+    """Return Harfbuzz-shaped text from input for font.
+
+    NOTE: copy-pasta from harfbuzz.py because fontbakery doesn't know how to import it.
```
chrissimpkins: Will you point me to the location of this file so that I can have a look at the license?

qa/check-fea.py
```diff
@@ -322,5 +328,132 @@ def com_google_fonts_check_googlesans_features_variable_italics(ttFont):
...
 profile.test_expected_checks(GOOGLESANS_PROFILE_CHECKS, exclusive=True)
+
+
+def shape_text(
+    font_path: str,
+    text: str,
+    script: str,
+    language: str,
+    features: Dict[str, bool],
+    variations: Optional[Dict[str, float]] = None,
```
chrissimpkins: define var font designspace slice with {'axis tag': float, 'axis tag': float} here?

qa/check-fea.py
```diff
@@ -322,5 +328,132 @@ def com_google_fonts_check_googlesans_features_variable_italics(ttFont):
...
 profile.auto_register(globals())
 profile.test_expected_checks(GOOGLESANS_PROFILE_CHECKS, exclusive=True)
+
+
+def shape_text(
+    font_path: str,
+    text: str,
+    script: str,
+    language: str,
+    features: Dict[str, bool],
```
chrissimpkins: `{'fea tag': True}` to activate optional features?

qa/check-fea.py
chrissimpkins: OK I see now. Yeah, the json files are divided up by script and feature.  I think that this runs once per font and tests against the first json file path returned by `shaping_basedir.glob("*.json")` then exits the loop when you yield the PASS/FAIL.

qa/check-fea.py
chrissimpkins: Got it, this is the TTFont fvar table `instance.coordinates` format: `{'opsz': 18.0, 'wght': 400.0}`

qa/check-fea.py
```diff
@@ -322,5 +328,132 @@ def com_google_fonts_check_googlesans_features_variable_italics(ttFont):
...
+            yield PASS, "No regression detected"
+        else:
+            yield FAIL, f"Expected and actual shaping not matching for {shaping_file}."
+
+
 profile.auto_register(globals())
 profile.test_expected_checks(GOOGLESANS_PROFILE_CHECKS, exclusive=True)
+
+
+def shape_text(
```
chrissimpkins: We need a direction parameter too to support Hebrew tests (and to future proof for other RTL scripts coming soon)?

qa/shaping/armenian_c2sc.json
```diff
@@ -0,0 +1,2659 @@
+{
+  "input": {
+    "text": [
+      "GS ĢȘ",
+      "ԵԳԶՋԷԾ,;"
+    ],
+    "features": {
+      "c2sc": true
+    },
+    "script": "armn",
```
chrissimpkins: This is the correct way to define script/lang in multi-script text runs? 🤔

qa/update_shaping_test_data.py
```diff
@@ -0,0 +1,116 @@
+from __future__ import annotations
+
+from pathlib import Path
+from typing import Any, Dict, List, Optional
+
+import uharfbuzz as hb
+from fontTools.ttLib import TTFont
+
+
+def shape_text(
```
chrissimpkins: Same comment as above in the fontbakery profile source.  Do we need a direction parameter for RTL scripts?

From [Harfbuzz docs](https://harfbuzz.github.io/setting-buffer-properties.html):

> Buffers also need to carry information about the script, language, and text direction of their contents. You can set these properties individually:

```cpp
      hb_buffer_set_direction(buf, HB_DIRECTION_LTR);
      hb_buffer_set_script(buf, HB_SCRIPT_LATIN);
      hb_buffer_set_language(buf, hb_language_from_string("en", -1));
```

requirements.txt
```diff
@@ -21,6 +21,7 @@ pytz==2020.1              # via fs
 six==1.15.0               # via fs
 ufo2ft[cffsubr]==2.15.0   # via fontmake
 ufolib2==0.8.0            # via fontmake, glyphslib
+uharfbuzz==0.14.0
```
chrissimpkins: Not a build dependency and we don't include fontbakery in the requirements.txt file.  Want to stick the uharfbuzz  install in the GH Actions configuration or pull it out into a requirements-dev.txt file?  We can add the requirements-dev.txt installs to a dev venv build Makefile target.

We want the fontbakery version to float to new releases in order to automatically pick up new tests, bug fixes.  How important is it that we pin the uharfbuzz version rather than taking latest for these tests?

chrissimpkins (changes requested):
This is a fantastic idea Nikolaus and Jany.  I love it.  I have a handful of comments inline.  The fontbakery json test file loop and direction parameter may need to be addressed.  Let me know what you think.

One other general comment. It looks like uharfbuzz is on HB v2.7.2 (HB v2.7.3 is in master branch since December but not released to PyPI yet) and harfbuzz is at v2.7.4. Let's discuss how important HB and uHB versions are in our testing.  It looks like uHB is doing some Python-y/Cython-y things that lead to releases off of the HB release cycle.  We'll either need to decide that we want to let the uHB version float to the latest version or add this to a maintainer todo list for a periodic maintenance version bump.  Unless there is a reason to pin the version, I suggest that we support the former.  If there is a reason to pin it, then I'll emphasize the comment about moving the dependency to a requirements-dev.txt file so that we can run pip-compile on the file and not have all of our build dependencies update at the same time. 

And lastly, we should make sure that someone plans to maintain uHB if we are going to build our shaping regression tests around the project.   @davelab6 @anthrotype

chrissimpkins:
We'll need to add maintainer documentation to `docs/MAINTAINER.md` also.  Can be done in a separate PR.  I'll get #120 merged first and then we can address it.

qa/check-fea.py
madig: Ah, that comment is stale. Basically I have to copy the code from `update_shaping_test_data.py` into `check-fea.py` because fontbakery trips over itself trying to import it.

qa/shaping/armenian_c2sc.json
madig: I don't know :D

requirements.txt
madig: Let's do a requirements-dev.txt. Can I add that to the `setup` Makefile target or is that just for production builds?

qa/update_shaping_test_data.py
madig: Will implement that then.

qa/check-fea.py
madig: Hm? It yields pass/fail for every json file for every font. Check with `fontbakery check-profile -c regression -C --loglevel PASS qa/check-fea.py build/GoogleSans/variable/expert/*.ttf build/GoogleSans/static/expert/*.ttf`.

qa/check-fea.py
madig: I now generate HTML diff pages. Very rough, but gives you an idea.

qa/check-fea.py
chrissimpkins: So nothing is coming from licensed source in an external project?  If so, we are ok here.

qa/check-fea.py
chrissimpkins: OK, sgtm. I was under the impression that fontbakery behaves differently. Thanks Nikolaus.

chrissimpkins:
@simoncozens Nikolaus is developing a uHarfbuzz based shaping regression test approach via shaping string diffs.  This will include a collection of "important" text runs that should be checked for shaping regressions as we update the sources with all of the new i18n design work.  Are you willing to review this PR and weigh in on with your thoughts about the approach?

qa/shaping/armenian_c2sc.json
chrissimpkins: Let's leave this unresolved for now.

requirements.txt
chrissimpkins: `setup` target is OK by me.  Thanks Nikolaus

chrissimpkins: Any thoughts about how to handle the uHB version? Pin or not?

qa/check-fea.py
chrissimpkins: > I now generate HTML diff pages. Very rough, but gives you an idea.

Are you thinking that we might run into massive diffs that are difficult to view in the GitHub UI CI output window?  If not, IMO the better approach would be to support the diff as unified diff text in the fontbakery CI output rather than requiring this to be run locally in order to understand what is different.  Or are you thinking that we push those HTML pages somewhere that allows them to be downloaded?

simoncozens:
Interesting - as you know I've been working on exactly the same thing! (https://github.com/simoncozens/nastaliq-engineering/blob/master/gnipahs.py)

I'll look in detail later but a few initial thoughts:

* This could be added to fontbakery; it's not something that's terribly GS-specific.
* JSON seems a bit heavy. If you serialise the buffer in a way that's compatible with hb-shape (see https://github.com/simoncozens/font-engineering/blob/49bede2d535bff8b35b634df4978fbc87561d726/vharfbuzz.py#L87-L97) then your test files are shorter and easier to read, and you have more flexibility in terms of generating the testing expectations.
* You don't need to pass direction or script just to get e.g. RTL support; `guess_segment_properties` does that.
* Features should take a feature string (for partial application), or at least allow for integers as well as booleans (alternates' tc.)
* Probably better to make the `hb.Face`/`hb.Font`/etc. a fixture rather than pulling in the font every time.

madig:
Thanks for reviewing! I agree that this would be nice to have in FB, my intention was to come up with something first and generalize later. I'd be fine with dropping this in favor of whatever you are working on if you target FB upstream.

simoncozens:
No, this is good and much more thorough than mine. OK, we'll polish it here then turn it into a FB PR.

qa/check-fea.py
```diff
@@ -322,5 +331,210 @@ def com_google_fonts_check_googlesans_features_variable_italics(ttFont):
...
+
+        try:
+            shaping_input = shaping_input_doc["input"]
+        except KeyError:
+            yield FAIL, (f"{shaping_file}: Must have an 'input' key dict.")
+            return
+        try:
+            shaping_texts = shaping_input["text"]
+            shaping_features = shaping_input["features"]
+            shaping_script = shaping_input["script"]
```
simoncozens: Suggest making these two optional, as Harfbuzz can guess them if they're not provided. (Normally they wouldn't be provided when shaping, unless you're testing `locl` stuff.)

qa/check-fea.py
```diff
@@ -322,5 +331,210 @@ def com_google_fonts_check_googlesans_features_variable_italics(ttFont):
...
+            yield FAIL, (f"{shaping_file}: Must have an 'output' key dict.")
+            return
+        try:
+            shaping_texts_expected = shaping_output[filename.name]
+        except KeyError:
+            yield FAIL, f"{shaping_file}: No entry found for {filename.name}"
+            return
+
+        if "fvar" in tt:
+            shaped_texts = shape_variable(
```
simoncozens: Avoid repetition here:

```
if "fvar" in tt:
    shape_function = shape_variable
else:
    shape_function = shape_static

shaped_text = shape_function(...)
```

chrissimpkins:
Thank you very much Simon. Much appreciated.  

> This could be added to fontbakery; it's not something that's terribly GS-specific.

Definitely agree. There are likely to also be generalizable shaping issue text runs across those that we will collect for the scripts in this project. It would be worth figuring out a way to export those out of this project so that others can use them with whatever approach is developed in fontbakery or elsewhere.  

For now we need an approach that gets the job done here to move ahead with script merges.  Let's continue the discussion about how to move it out of the GS repo.

qa/check-fea.py
```diff
@@ -322,5 +331,210 @@ def com_google_fonts_check_googlesans_features_variable_italics(ttFont):
...
+                shaping_language,
+                shaping_direction,
+                shaping_features,
+                shaping_comparison_mode,
+            )
+
+        if shaped_texts == shaping_texts_expected:
+            yield PASS, f"{shaping_file}: No regression detected"
+        else:
+            shaped_texts_json = json.dumps(shaped_texts, indent=2).split("\n")
```
simoncozens: Serializing the buffer to hb-shape-like strings would make this tidier and easier for end-user to read.

qa/check-fea.py
```diff
@@ -322,5 +331,210 @@ def com_google_fonts_check_googlesans_features_variable_italics(ttFont):
...
+        fontdata = fontfile.read()
+
+    face = hb.Face(fontdata)
+    font = hb.Font(face)
+    upem = face.upem
+    if variations is not None:
+        font.set_variations(variations)
+
+    font.scale = (upem, upem)
+    hb.ot_font_set_funcs(font)
```
simoncozens: Suggest making this a FontBakery condition rather than re-setting it up for each string.

qa/check-fea.py
```diff
@@ -322,5 +331,210 @@ def com_google_fonts_check_googlesans_features_variable_italics(ttFont):
...
+            }
+            for info, pos in zip(infos, positions)
+        ]
+    elif shaping_comparison_mode is ComparisonMode.GLYPHSTREAM:
+        return [font.get_glyph_name(info.codepoint) for info in infos]
+    else:
+        raise ValueError(f"Unknown comparison mode {shaping_comparison_mode}.")
+
+
+def shape_variable(
```
simoncozens: I'm really suspicious of the repetition in these two functions. They're basically the same, and most of the code is taken up with unpacking parameter lists and packing them up again. Maybe you should be passing a parameters hash initially, or maybe you should be making more use of `kwargs`.

qa/check-fea.py
```diff
@@ -322,5 +331,210 @@ def com_google_fonts_check_googlesans_features_variable_italics(ttFont):
...
+                f"{shaping_file}: Expected and actual shaping not matching. "
+                f"Open {f.name} in your browser for details."
+            )
+
+
 profile.auto_register(globals())
 profile.test_expected_checks(GOOGLESANS_PROFILE_CHECKS, exclusive=True)
+
+
+# XXX: Below is a copy-pasta of update_shaping_test_data.py because I can't
```
simoncozens: That seems like a problem worth solving. There should be a generic shaping library (see my Vharfbuzz), which both this and the `update_shaping_test_data` uses.

simoncozens (changes requested):

qa/check-fea.py
madig: Cosimo suggested requiring them to rule out HB misguessing.

simoncozens: HB guessing wrong is an argument for *allowing* them, but you only need to supply them when HB *does* guess wrong. (But does it really guess wrong? Do we have examples of this happening?)

qa/check-fea.py
madig: The problem is that doing `from .update_shaping_test_data import *` results in FB keeling over:

```
Traceback (most recent call last):
  File "/home/nikolaus/googlesans/venv/bin/fontbakery", line 8, in <module>
    sys.exit(main())
  File "/home/nikolaus/googlesans/venv/lib/python3.8/site-packages/fontbakery/cli.py", line 21, in main
    runpy.run_module(
  File "/usr/lib/python3.8/runpy.py", line 210, in run_module
    return _run_code(code, {}, init_globals, run_name, mod_spec)
  File "/usr/lib/python3.8/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/home/nikolaus/googlesans/venv/lib/python3.8/site-packages/fontbakery/commands/check_profile.py", line 369, in <module>
    sys.exit(main())
  File "/home/nikolaus/googlesans/venv/lib/python3.8/site-packages/fontbakery/commands/check_profile.py", line 246, in main
    profile = get_profile()
  File "/home/nikolaus/googlesans/venv/lib/python3.8/site-packages/fontbakery/commands/check_profile.py", line 231, in get_profile
    imported = get_module(args.profile)
  File "/home/nikolaus/googlesans/venv/lib/python3.8/site-packages/fontbakery/commands/check_profile.py", line 213, in get_module
    imported = get_module_from_file(name)
  File "/home/nikolaus/googlesans/venv/lib/python3.8/site-packages/fontbakery/checkrunner.py", line 1916, in get_module_from_file
    module_spec.loader.exec_module(module)
  File "<frozen importlib._bootstrap_external>", line 783, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "qa/check-fea.py", line 426, in <module>
    from .update_shaping_test_data import *
ModuleNotFoundError: No module named 'fontbakery.__file_modules'
```

Haven't had the nerve to dive into FB yet.

simoncozens: OK, looks like FB has a problem with relative imports. Importing from "qa.update_shaping_test_data" (and setting PYTHONPATH) works fine. I'll raise a FB issue.

qa/check-fea.py
madig: Golfed them a bit.

qa/check-fea.py
simoncozens: https://github.com/googlefonts/fontbakery/issues/3161

simoncozens:
> Is uHB going to be maintained?

Yes. It's considered part of the Harfbuzz project.

> who is the primary maintainer?

Just, although the Harfbuzz maintainers can also commit to master.

> is it the best project to base our shaping regression strategy on?

Sure, what else is there? And now if it ever goes wrong, you can replace it with command-line call-outs to `hb-shape`.

qa/check-fea.py
madig: With `from qa.update_shaping_test_data import ...` I get `ModuleNotFoundError: No module named 'qa'` :(

qa/check-fea.py
madig: I have zero idea! I am however obsessed with edge cases. But can make optional, as well as direction.

qa/check-fea.py
simoncozens: `PYTHONPATH=.` (or in the Makefile `export PYTHONPATH=.`)

madig: Or inside the .py file actually :O ugly but works.

chrissimpkins:
>> is it the best project to base our shaping regression strategy on?

> Sure, what else is there? And now if it ever goes wrong, you can replace it with command-line call-outs to hb-shape.

Harfbuzz :)

simoncozens:
uHarfbuzz is the python wrapper around Harfbuzz. If you want to use Harfbuzz from Python, this is how you do it.

qa/check-fea.py
madig: I put `hb.Face` into a condition now. Is `hb.Font` that other stuff a heavy op?

chrissimpkins:
> uHarfbuzz is the python wrapper around Harfbuzz. If you want to use Harfbuzz from Python, this is how you do it.

I think that this conversation originated from these statements that I made in https://github.com/googlefonts/googlesans/pull/137#pullrequestreview-580920035

> One other general comment. It looks like uharfbuzz is on HB v2.7.2 (HB v2.7.3 is in master branch since December but not released to PyPI yet) and harfbuzz is at v2.7.4. Let's discuss how important HB and uHB versions are in our testing. It looks like uHB is doing some Python-y/Cython-y things that lead to releases off of the HB release cycle. 

By "is this the best project", I meant the uHarfbuzz Python wrapper for Harfbuzz.  An alternative approach is to do all of this with Harfbuzz/ the hb-shape utility without accessing it via Python or using it in fontbakery.  I was thinking out loud.  I think that these questions remain.  How much does uHarfbuzz diverge from Harfbuzz now and in the future, is uHarfbuzz maintained on a release schedule (relative to Harfbuzz) that this remains a reliable approach for shaping regression testing in this project down the road? Do Harfbuzz versions really matter for this style of testing and how do we approach the version of the test dependency that we use?

madig:
I think the big draw with a Python wheel is that we can just `pip install` it anywhere without messing with compilers. We don't seem to be missing too much when staying on 2.7.2, but maybe we can budget some time for uHB maintenance? I see it's still on Travis and Appveyor, and could also do with some typing stubs. Typical housekeeping stuff.

madig:
(Awaiting further reviews. Or are the unit tests ready to go after we decided on HB or uHB?)

requirements.txt
madig: Hm. I think we can leave unpinned and revisit if problems crop up.

madig:
Oh wait. generating HTML diffs and writing them to temp is ok locally but problematic on the CI... except maybe if they are uploaded as artifacts?

simoncozens:
This is why I suggested using the hb-shape strings and diffing those: you can do it as a textual diff, and add that to the normal FB report.

madig:
Hm. I worry about readability on the command line. Well, I'll try something and see if it's acceptable.

simoncozens:
If the tests pass, it'll be readable. If the tests don't pass, you don't *want* readable, you want detailed enough to diagnose and fix.

> We don't seem to be missing too much when staying on 2.7.2, but maybe we can budget some time for uHB maintenance?
> I see it's still on Travis and Appveyor, and could also do with some typing stubs.

Moving to GitHub actions would be good; I think if that's done, we can work on automatically sync'ing the Harfbuzz versions and doing. a new release every time HB is released. (See https://github.com/harfbuzz/harfbuzz/issues/2839) That would cut the maintenance burden considerably.

madig:
Report looks like this now:

```
 >> com.google.fonts/check/googlesans/features/regression
    But does it shape?
    with build/GoogleSans/variable/expert/GoogleSans-Italic[opsz,wght].ttf

    * FAIL: qa/shaping/armenian_case.json: Expected and actual shaping not matching.
        Expected for opsz=18.0,wght=400.0:
          uni055A.BRACKET.18=0+228,0|uni053B=1+637,0|space=2+224,0|Ech_Vewarm.liga=3+1282,0|uni055D=4+255,0|space=5+224,0|uni0544=6+665,0|uni0589.case=7+239,0
          Men_Nowarm.liga=0+1331,0|space=1+224,0|Men_Echarm.liga=2+1398,0|space=3+224,0|Men_Iniarm.liga=4+1381,0|space=5+224,0|Vew_Nowarm.liga=6+1294,0|space=7+224,0|Men_Xeharm.liga=8+1601,0
          Ech_Vewarm.liga=0+1312,0|uni055B.comb=1@-1097,0|space=2+224,0|Men_Echarm.liga=3+1398,0|uni055B.comb=4@-398,0|space=5+224,0|Men_Iniarm.liga=6+1381,0|uni055B.short.comb=7@-388,0
        Actual:
          uni055A.BRACKET.18=0+228,0|uni053B=1+637,0|space=2+224,0|Ech_Vewarm.liga=3+1282,0|uni055D=4+255,0|space=5+224,0|uni0544=6+665,0|uni0589.case=7+239,0
          Men_Nowarm.liga=0+1331,0|space=1+224,0|Men_Echarm.liga=2+1398,0|space=3+224,0|Men_Iniarm.liga=4+1381,0|space=5+224,0|Vew_Nowarm.liga=6+1294,0|space=7+224,0|aMen_Xeharm.liga=8+1601,0
          Ech_Vewarm.liga=0+1312,0|uni055B.comb=1@-1097,0|space=2+224,0|Men_Echarm.liga=3+1398,0|uni055B.comb=4@-398,0|space=5+224,0|Men_Iniarm.liga=6+1381,0|uni055B.short.comb=7@-388,0
    * FAIL: qa/shaping/armenian_case.json: Expected and actual shaping not matching.
        Expected for opsz=17.0,wght=400.0:
          uni055A.case=0+292,0|uni053B=1+640,0|space=2+256,0|Ech_Vewarm.liga=3+1347,0|uni055D=4+254,0|space=5+256,0|uni0544=6+705,0|uni0589.case=7+259,0
          Men_Nowarm.liga=0+1352,0|space=1+256,0|Men_Echarm.liga=2+1476,0|space=3+256,0|Men_Iniarm.liga=4+1425,0|space=5+256,0|Vew_Nowarm.liga=6+1357,0|space=7+256,0|Men_Xeharm.liga=8+1674,0
          Ech_Vewarm.liga=0+1427,0|uni055B.comb=1@-1213,0|space=2+256,0|Men_Echarm.liga=3+1476,0|uni055B.comb=4@-477,0|space=5+256,0|Men_Iniarm.liga=6+1425,0|uni055B.short.comb=7@-401,0
        Actual:
          uni055A.case=0+292,0|uni053B=1+640,0|space=2+256,0|Ech_Vewarm.liga=3+1347,0|uni055D=4+254,0|space=5+256,0|uni0544=6+705,0|uni0589.case=7+259,0
          Men_Nowarm.liga=0+1352,0|sapace=1+256,0|Men_Echarm.liga=2+1476,0|space=3+256,0|Men_Iniarm.liga=4+1425,0|space=5+256,0|Vew_Nowarm.liga=6+1357,0|space=7+256,0|Men_Xeharm.liga=8+1674,0
          Ech_Vewarm.liga=0+1427,0|uni055B.comb=1@-1213,0|space=2+256,0|Men_Echarm.liga=3+1476,0|uni055B.comb=4@-477,0|space=5+256,0|Men_Iniarm.liga=6+1425,0|uni055B.short.comb=7@-401,0


    Result: FAIL

 >> com.google.fonts/check/googlesans/features/regression
    But does it shape?
    with build/GoogleSans/static/expert/GoogleSans-Regular.ttf

    * FAIL: qa/shaping/armenian_case.json: Expected and actual shaping not matching.
        Expected:
          uni055A.case=0+224,0|uni053B=1+637,0|space=2+232,0|Ech_Vewarm.liga=3+1280,0|uni055D=4+255,0|space=5+232,0|uni0544=6+664,0|uni0589.case=7+238,0
          Men_Nowarm.liga=0+1331,0|space=1+232,0|Men_Echarm.liga=2+1397,0|space=3+232,0|Men_Iniarm.liga=4+1381,0|space=5+232,0|Vew_Nowarm.liga=6+1295,0|space=7+232,0|Men_Xeharm.liga=8+1603,0
          Ech_Vewarm.liga=0+1310,0|uni055B.comb=1@-1098,0|space=2+232,0|Men_Echarm.liga=3+1397,0|uni055B.comb=4@-400,0|space=5+232,0|Men_Iniarm.liga=6+1381,0|uni055B.short.comb=7@-387,0
        Actual:
          uni055A.case=0+224,0|uni053B=1+637,0|space=2+232,0|Ech_Vewarm.liga=3+1280,0|uni055D=4+255,0|space=5+232,0|uni0544=6+664,0|uni0589.case=7+238,0
          Men_Nowarm.liga=0+1331,0|space=1+232,0|Men_Echarm.liga=2+1397,0|aspace=3+232,0|Men_Iniarm.liga=4+1381,0|space=5+232,0|Vew_Nowarm.liga=6+1295,0|space=7+232,0|Men_Xeharm.liga=8+1603,0
          Ech_Vewarm.liga=0+1310,0|uni055B.comb=1@-1098,0|space=2+232,0|Men_Echarm.liga=3+1397,0|uni055B.comb=4@-400,0|space=5+232,0|Men_Iniarm.liga=6+1381,0|uni055B.short.comb=7@-387,0


    Result: FAIL
```

madig:
What remains to be done here? I can clean up the commits and split the PR up, the core code can go to master and the amrn tests can be appended to the Armenian PR.

chrissimpkins:
> If the tests don't pass, you don't want readable, you want detailed enough to diagnose and fix.

I'd argue that it would be *ideal* to have it show exactly what differs rather than requiring a second step to pull out a block of text strings for a local diff.  But perfect is the enemy of good.  Let's get something in place that does the job and we can update further as we go along.  No need to get bogged down on making this a perfect solution right now.  The full observed / expected output LGTM.  And I really like that it is in the CI rather than HTML output that must be dumped locally.

chrissimpkins:
> What remains to be done here? I can clean up the commits and split the PR up, the core code can go to master and the amrn tests can be appended to the Armenian PR.

@simoncozens is there anything else that you'd like to see to wrap things up here?

simoncozens:
Looks good to me!

chrissimpkins:
> Looks good to me!

Thank you very much for all of your input Simon!  Much appreciated.

chrissimpkins:
The new shaping regression testing source was opened in https://github.com/googlefonts/googlesans/pull/140 where it is separated from the Armenian test text runs here.

madig:
I have moved the test code to https://github.com/googlefonts/googlesans/pull/140 and added the test files to https://github.com/googlefonts/googlesans/pull/129. This PR is hereby superseded.

Issue #135, opened 2021-01-21
Approach to optical size 18 glyph substitutions currently defined with OT rvrn feature (Glyphs "bracket layers")


chrissimpkins:
Nikolaus proposed a transition from the `rvrn` to `rclt` feature to extend optical size dependent glyph substitution support to glyphs that require other feature substitutions.  This issue arose during the merge of the Armenian work in #129 and the discussion began at https://github.com/googlefonts/googlesans/pull/129#discussion_r561889791.

_Let's define the behavior here as **optical size dependent substitution = OSDS** since the "Glyphs bracket layer", "rvrn", "rclt" terminology is application/source format and implementation dependent, and confuses the issue_.

## Issue

From @madig (https://github.com/googlefonts/googlesans/pull/129#discussion_r562023481):

> All bracket glyphs for glyphs swapped in by other features. The problem has been expounded in fonttools/fonttools#1371.
>
>The default behavior in VFs is to use the rvrn feature, which is applied first thing. This is unfortunately incompatible with having bracket glyphs for glpyhs that are swapped in by other features. Take "g", "g.sc" and "g.sc.BRACKET.18" as an example:
>
>In opsz == 17, i.e. for Text, turning the smcp option on results in "g" being replaced by "g.sc". That's as intended.
In opsz == 18, i.e. for Display, you want the end result to be "g.sc.BRACKET.18". However: rvrn runs first ("g.sc" -> "g.sc.BRACKET.18") and finds no "g.sc" to replace (smcp runs later!). It does nothing. Then, the smcp feature runs and replaces "g" by "g.sc". That's your end result rather than "g.sc.BRACKET.18" simply because rvrn ran too early.
As the linked issue explains, there's no easy way out. rclt is not supported by Adobe apps and to make it mostly work with rvrn, you have to, I think, add bracket glyphs for all parent glyphs of bracketed glyphs (i.e. here "g.BRACKET.18") and have rvrn switch that out and add an entry in scmp to replace g.BRACKET.18 with g.sc.BRACKET.18. Put another way, you have to double all your subs for bracket glyphs (e.g. fonttools/fonttools#1371 (comment)).

From https://github.com/googlefonts/googlesans/pull/129#discussion_r561926372:

> AFAIK, rclt is the only sensible way of getting bracket glyphs to work if you want to stack features (it applies substitutions last, rvrn is applied before other features are applied).

[...]

PR #139, opened 2021-02-01
rvrn salvation proof of concept

madig:
DO NOT MERGE

I think this solves the "g" and "G" woes?

I took out `<sub name="g.sc.ss04" with="g.sc.ss04.BRACKET.18"/>` because I think that's wrong, the target should be the lowercase variant... which does not exist.

Feature files centralized outside the UFOs to make sweeping changes easy.


chrissimpkins:
> I took out <sub name="g.sc.ss04" with="g.sc.ss04.BRACKET.18"/> because I think that's wrong, the target should be the lowercase variant... which does not exist.

See https://github.com/googlefonts/googlesans/issues/135#issuecomment-771193168. Let's discuss whether we need to support these ss04 rvrn subs at all.  Waiting for feedback from Edd Harrington.

source/GoogleSans/GoogleSans-Bold.ufo/glyphs/contents.plist
```diff
@@ -2786,6 +2786,8 @@
     <string>g.glif</string>
     <key>g.logo</key>
     <string>g.logo.glif</string>
+    <key>g.rvrn</key>
```
chrissimpkins, line 2801: I realize that this is just a test.  Jany mentioned that we will need to keep these in the `[glyphname].[+/- feature tag].BRACKET.18` format to support the trips through glyphsLib.  FWIW, I prefer the .rvrn extensions here :)

source/GoogleSans/GoogleSans-Bold.ufo/glyphs/g.rvrn.glif
```diff
@@ -0,0 +1,7 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<glyph name="g.rvrn" format="2">
+  <advance width="624"/>
+  <outline>
+    <component base="g"/>
```
chrissimpkins, line 5: Nice. I like the use of components for these duplicates.  👍

chrissimpkins (changes requested):
I think that what you have is correct Nikolaus.

I went through the glyph designs in the GS-Regular and GST-Regular masters.  The behavior that we want is the following:

### smcp off

`g` opsz 17 master design interpolates to `g` opsz 18 master design.  There is no rvrn issue here.  Nothing needs to change in the `g` designs.

### smcp on

This is where things get confusing.  There are four different g.sc forms across the 17 and 18 opsz masters:

- g.sc @ opsz 18 master
- g.sc.BRACKET.18 @ opsz 18 master
- g.sc @ opsz 17 master
- g.sc.BRACKET.18 @ opsz 17 master

The interpolation behavior should follow the form changes that happen with the `G`.  Spurred form @ opsz 17 interpolation to sub-18 wider form, followed by abrupt transition to non-spurred form @ opsz 18 (rvrn transition).

upper-G

But here is where we are with the `g.sc` over the opsz axis now:

g sc-rvrn

The interpolation between opsz 17 g.sc to opsz 18 g.sc happens, but the rvrn transition is missing.  We need to have the transition to the `g.sc.BRACKET.18` outline definition in the max opsz master.  [Aside: it is not clear to me if/how the min opsz master version of `g.sc.BRACKET.18` plays into this, maybe this can be removed?]

I think that what we need are the following glyph definitions:

- `g` (opsz 17 master) - no change
- `g` (opsz 18 master) - no change
- `g.sc` (opsz 17 master) - no change, this is the min end of the smcp form interpolation range
- `g.sc` (opsz 18 master) - no change, this is the max end of the smcp form interpolation range just prior to the rvrn transition to the `g.sc.BRACKET.18` opsz 18 form when opsz is defined at 18+
- `g.BRACKET.18`  (opsz 17 master) - composite with opsz 17 master `g` component (= rvrn when smcp is not active)
- `g.BRACKET.18` (opsz 18 master) - composite with opsz 18 master `g` component (= rvrn when smcp is not active)
- `g.sc.BRACKET.18` (opsz 17 master) - `g.sc.BRACKET.18` currently defined in the opsz 18 master
- `g.sc.BRACKET.18` (opsz 18 master) - `g.sc.BRACKET.18` currently defined in the opsz 18 master

The new designspace rvrn sub is then:

```xml
<sub name="g" with="g.BRACKET.18"/>
```

and features.fea needs to define:

```fea
# this defines the smcp sub at opsz 17
sub g by g.sc;

# this defines the smcp sub at opsz 18
sub g.BRACKET.18 by g.sc.BRACKET.18;
```

Whew.

I believe that the only thing that we need to do is change the `g.rvrn` glyphs to `g.BRACKET.18` so that we support the format used in glyphsLib with future glyphs bracket layer imports.  Maybe also exports of these rvrn definitions back into glyphs source land too? Not sure how the round trip works for the rvrn definitions.  Ideally these show up as new [18] bracket layers on opsz 17 and opsz18 master `g` in Glyphs.app and turn back into `g.BRACKET.18` in our glyphs -> UFO transform on the trip back in.

chrissimpkins:
[Your build ](https://github.com/googlefonts/googlesans/actions/runs/528458601) is a thing of beauty :) smcp forms shown below.

g-sc-fixed

There is a minor snap in the advance width during the `g` opsz 17 master to `g` opsz 18 master interpolation -> substitution (not smcp, not shown in visual above) as you hit a definition of opsz 18.  It looks like we introduce this with the new rvrn on `g` to `g.BRACKET.18`/`g.rvrn` because the adv width on the two designs differs and we don't quite get close enough to the adv width on the opsz 18 form with interpolation before the substitution happens.  We could probably iron this out a bit with avar but I don't think that it is an issue for us.  There should be no animated interpolation between 17 and 18.  I think that this is OK.

source/GoogleSans/GoogleSans-Bold.ufo/glyphs/contents.plist
madig: Hm, does Glyphs.app matter here? Maybe for the preview 🤔

chrissimpkins: That is where design vendors are.  And the Armenian process suggests that there may be a few round trips back and forth to Glyphs.app editors.

madig:
Do I understand correctly that my change is what you propose should happen? Your review implies I need to change something?

chrissimpkins:
> Do I understand correctly that my change is what you propose is what should happen? Your review implies I need to change something?

The behavior is correct with changes here.  I think that we need to discuss glyph names further.  I like the format `[glyphname].rvrn` but I don't think that is supported automatically on glyphs -> ufo imports via glyphsLib, nor on exports of production source back to glyphs format for change requests from vendors (and I suppose new initial branches from the upstream prod sources). They can be forced to work in UFO only after they initially push their glyphs files or start new changes, but my understanding was that we intended to support full round trips between UFO <-> glyphs.  Use of `.rvrn` means we need to edit glyphsLib imported glyph names on our side because they come in as `*.BRACKET.18` from the glyphs [18] layer definitions. 

We do introduce a new spacing issue in the lowercase `g` opsz interpolation that I think is minor and acceptable.

madig:
For the `y`, there already is a `y.BRACKET.18` so I'll have to sneak an rvrn in there...

Edit: actually I'll try to make everything work with `.rvrn` first.

madig:
Grrr. Where do I hack rvrn into `Gcommaaccent.alt`

chrissimpkins:
I started a spreadsheet with required glyphs here https://docs.google.com/spreadsheets/d/1oP9ZkiLpTDw7N2RtbaIbYkOEHSfXLHYgQKSAaJzBGCo/edit?resourcekey=0-0IdeRO6xWXFXdY5KUgoM1w#gid=0

I think that we can do the glyph gymnastics similar to the `.sc` approach and then add the new opsz 18 substitution in `aalt`?

madig:
Thanks, will try.

Also, there's Gdotaccent, Gdotaccent.alt and Gdotaccent.BRACKET.18 but no Gdotaccent.alt.BRACKET.18? And Gdotaccent and Gdotaccent.alt look the same?

Also also, there's a `sub Tcommaaccent by tcommaaccent.sc` in `smcp` so there should maybe be a `tcommaaccent.sc.BRACKET.18`? Same for gcommaaccent.sc, kcommaaccent.sc, rcommaaccent.sc. And they all may need bracket variants as well?

Also also also, I see e.g. `sub tcedilla by tcommaaccent` in `locl`. I suppose we now need a `tcedilla.rvrn` plus `sub tcedilla.rvrn by tcommaaccent.BRACKET.18`.

chrissimpkins:
🤯 

I will go through all of the designs and see if I can figure out what was intended.  I likely won't get to this before the weekend.  I'll be tracking down someone to weigh in on whether we need to keep the `.ss04` set today.  More to come in this thread.

madig:
🤯 indeed. I'll try to get the Latin stuff up shortly. Armenian may be more tricky due to the way the features are written.

madig:
Oh btw, `glogoligature` does not trigger the `liga` transformation in opsz=18 because we now have `g.rvrn lo g.rvrn oli g.rvrn ature`. More stuff. I guess this ideal fodder for regression testing. The ligas also don't trigger when ss04 is active, but maybe that's out of scope.

madig:
Currently fighting the Google.logo `liga`s. Why are there two glyphs `Google.logo` and `Googlelogo`  that are both decomposed outlines? Same for the single-letter `.logo` glyphs.

chrissimpkins:
> I'll be tracking down someone to weigh in on whether we need to keep the `.ss04` set today. More to come in this thread.

https://github.com/googlefonts/googlesans/issues/135#issuecomment-773450365

chrissimpkins:
> Oh btw, `glogoligature` does not trigger the `liga` transformation in opsz=18 because we now have `g.rvrn lo g.rvrn oli g.rvrn ature`. More stuff. I guess this ideal fodder for regression testing.

`liga` regression tests #170 

`calt` regression tests #165 

Both features have logo forms with lowercase `g`.

chrissimpkins:
We will need to review the smcp set.  Based on the VF hb shaping data that was added in #190, there may be other  *.BRACKET.18 glyph subs that are not properly defined as the sc forms with an active smcp feature.

chrissimpkins:
OK to close?

madig:
I think so!

PR #140, opened 2021-02-03
Add text shaping regression test suite

madig:
Split off from https://github.com/googlefonts/googlesans/pull/137. Contains no tests yet.

This is meant to be a good-enough-for-now implementation that should later be streamlined for upstreaming into FontBakery.

### DIFF
--- a/.github/workflows/static-artifact-ci.yml
+++ b/.github/workflows/static-artifact-ci.yml
@@ -29,11 +29,11 @@ jobs:
         with:
           python-version: ${{ steps.config.outputs.py-version }}
       - name: Python build dependency cache lookup
-        uses: actions/cache@v1
+        uses: actions/cache@v2
         with:
           path: ~/.cache/pip
           # Check for requirements file cache hit
-          key: ${{ runner.os }}-pip-${{ hashFiles('${{ steps.config.outputs.dependpath }}') }}
+          key: ${{ runner.os }}-pip-${{ hashFiles('requirements*.txt') }}
       - name: Install Python build dependencies
         uses: py-actions/py-dependency-install@v2
         with:
--- a/.github/workflows/static-artifact-ci.yml
+++ b/.github/workflows/static-artifact-ci.yml
@@ -39,6 +39,7 @@ jobs:
         with:
           update-wheel: "true"
           update-setuptools: "true"
+          path: "requirements-dev.txt"
       - name: Build fonts
         run: ${{ steps.config.outputs.buildcommand }}
       - name: Font Bakery GS profile tests (Expert artifacts)
--- a/docs/MAINTAINER.md
+++ b/docs/MAINTAINER.md
@@ -248,3 +248,48 @@ The same as the [Glyphs.app workflow](#importing-a-glyphsapp-file), except we do
 Relies on testing material provided by vendors. Testing material should serve as a reference for how text should look and how features work. For testing, the vendor provided font is swapped with the merged font; if everything stays the same, the test is considered passed.
 
 ![Quality Assurance Workflow](assets/qa_process.png)
+
+### Adding and Updating Text Shaping Regression Test Files
+
+To ensure adding and changing feature code does not break existing font functionality, text shaping comparisons are run using HarfBuzz.
+
+The directory `qa/shaping/` contains `.json` files of the following format:
+
+```json
+{
+  "input": {
+    // Required, a list of strings to shape.
+    "text": [
+      "GS ĢȘ",
+      "ԵԳԶՋԷԾ,;"
+    ],
+    // Required, a dictionary of features to enable (true) or disable (false).
+    // Can be empty.
+    "features": {
+      "c2sc": true,
+      "ss04": true
+    },
+    // Optional, the ISO 15924 script tag for the text.
+    "script": "armn",
+    // Optional, the BCP 47 language tag for the text.
+    "language": "hy",
+    // Optional, the script direction, one of "ltr", "rtl", "ttb" or "btt".
+    "direction": "ltr",
+    // Optional, the shaping output. Either "full", which includes offsets and
+    // advance widths, or "glyphstream" for just the glyph names.
+    "comparison_mode": "full"
+  },
+  "output": {
+      // Per font file output goes here...
+  }
+}
+```
+
+The files currently have to be created by hand. Give them a descriptive name, ideally prefixed by the script they pertain to. To fill them with the shaping results of a list of fonts, essentially setting their output in stone, run:
+
+```
+$ python3 qa/update_shaping_test_data.py qa/shaping/my_file.json \
+    build/GoogleSans/variable/expert/*.ttf build/GoogleSans/static/expert/*.ttf
+```
+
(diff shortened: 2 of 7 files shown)


docs/MAINTAINER.md
```diff
@@ -248,3 +248,48 @@ The same as the [Glyphs.app workflow](#importing-a-glyphsapp-file), except we do
 Relies on testing material provided by vendors. Testing material should serve as a reference for how text should look and how features work. For testing, the vendor provided font is swapped with the merged font; if everything stays the same, the test is considered passed.
 
 ![Quality Assurance Workflow](assets/qa_process.png)
+
+### Adding and Updating Text Shaping Regression Test Files
```
chrissimpkins, line 252: ❤️

.github/workflows/static-artifact-ci.yml
```diff
@@ -39,6 +39,7 @@ jobs:
         with:
           update-wheel: "true"
           update-setuptools: "true"
+          path: "requirements-dev.txt"
```
chrissimpkins, line 42: Hmm... Not sure how our CI dependency caching will work with the new approach.  Here is where it is defined based on the requirements.txt file:

https://github.com/googlefonts/googlesans/blob/f583bf5a714aa5965c24a740289bbb69b23c5ded/.github/workflows/static-artifact-ci.yml#L36

There is a similar line in the variable font CI configuration.  I *think* that as long as we keep all build dependencies in the requirements.txt we are ok.

madig: Maybe it works perfectly if we just hash all the reqs 🤷‍♀️ Also the docs say cache@v1 can only hash directories?! https://docs.github.com/en/actions/guides/caching-dependencies-to-speed-up-workflows#input-parameters-for-the-cache-action

chrissimpkins: Like it!  Mind pushing one more commit so that we can see if it uses the cached versions after the run on https://github.com/googlefonts/googlesans/pull/140/commits/10facd03dade631c595616e3e2dd8a0f28c1ab90?

chrissimpkins:
Do you want to merge this as is or begin adding test text runs in this PR?

chrissimpkins:
Note to self: Let's Apache license the text run data files and make sure that the fontbakery custom profiles have Apache license in header or documented separately in repo.

madig:
I'd merge as is and immediately use in my `rvrn` battle.

.github/workflows/static-artifact-ci.yml
madig: Re-running jobs, maybe that works, too.

madig: Seems to work: https://github.com/googlefonts/googlesans/pull/140/checks?check_run_id=1831119026#step:6:67

madig:
@chrissimpkins anything else? I need this 😬

.github/workflows/static-artifact-ci.yml
chrissimpkins: 👍 ty!

chrissimpkins:
> @chrissimpkins anything else? I need this 😬

LGTM.  Apache licenses are all in place on the custom profiles.  GTG.

chrissimpkins:
changelog 647bdb31ea975816088bec1d18b5ec978532df01

madig:
Thanks! Picking apart the Armn PR next, then back to `rvrn`. Maybe I need to do a refactoring of the feature files first, to make fixing things across fonts less painful...

chrissimpkins:
I'll begin collecting text runs that we can use for the regression tests in the production scripts.  That seems to be one of the main issues blocking the merge of a good deal of work.  Now that we have the test suite we can dive into filling it out with a set of checks.   I love that we broke the logo ligature btw.  Lol and ugh...

madig:
I don't think we broke it. It more likely never worked outside of what the designers saw in Glyphs.app. 😬

simoncozens:
@chrissimpkins: Any objection to me talking the shaping part and expanding it? I have some ideas how we could make it test things more broadly in line with what I've been doing for Nastaliq/Myanmar. (By which I mean: using `stringbrewer` to generate test cases procedurally, testing for configurable "invalid" glyphs such as .notdef and visible virama, running collidoscope for collision testing.)

There's an Apache license on the top of the file - OK to develop in public or would you prefer it in a private repo for now?

chrissimpkins:
No objections at all Simon.

> There's an Apache license on the top of the file - OK to develop in public or would you prefer it in a private repo for now?

This is intentional for all of the scripts in this project.  Please do improve and feel free to make the source public.  Let us know when you have a repository so that we can follow and pitch in where possible.  Look forward to seeing what you come up with.

chrissimpkins:
The comment in https://github.com/googlefonts/googlesans/pull/140#issuecomment-781422481 applies to the build workflow (Makefiles, Py scripts), all GH Action configuration files, and other infrastructure scripts for anyone else who happens to be following this repo.  Please do take these files, adapt/improve, and let us know when you come up with something better so that we can use it here.  There should be an explicit license declaration in the header for files where this applies. If there is not, please let me know.

madig:
A font shaping fuzzer. I like it.

Please also invent a glyphset consistency checker that answers the question "If I have a `rvrn` substitution for my glyph `G`, which additional `rvrn` glyphs do I have to introduce to cover all the glyphs that `G` can be substituted into across the entire font, taking into account (de)composition?".

Potential future explorations could be making a machine-proofable model, which you give your scripts, languages and feature set and it spits out the glyphset plus the feature code to implement it all in a verified and conflict-free way. That would make shaping testing redundant.

simoncozens:
OK, I have gone a bit wild with this and it now bears hardly any resemblance to the Google Sans shaping tester, (the JSON format is quite different) but it does a few cool things:

* [code](https://github.com/simoncozens/font-engineering/blob/master/fontbakery-shaping.py)
* [example test suites](https://github.com/simoncozens/font-engineering/tree/master/qa/shaping_tests)
  * `basic-regression.json` is essentially what we have at the moment.
  * `forbidden-glyphs.json` tests to see if any bad glyphs turn up while shaping. Note that it can take [StringBrewer](https://pypi.org/project/stringbrewer/) patterns as well as plain strings.
  * `collisions.json` checks for colliding glyphs according to Collidoscope rules. Again, it takes both plain strings StringBrewer patterns.

Running `fontbakery check-profile ./fontbakery-shaping.py GoogleSans-Regular.ttf` turns up:

```
    * FAIL: qa/shaping_tests/forbidden-glyphs.json: Forbidden glyphs found while shaping.
      * 禁止 produced '.notdef'
       .notdef|.notdef
```

(which is fine, we should expect that)

but also:

```
    * FAIL: qa/shaping_tests/collisions.json: Collisions found while shaping.
      * icircumflex/icircumflex collision found in e.g. 'dîîd'
```

A kern needed, perhaps? :-)

The automated forbidden glyphs and collision tests are *much* more useful when it comes to complex non-Latins, where weird combinations may turn up visible viramas, clashes, etc.

PR #141, opened 2021-02-03
Greek fix for components issue

irenevlachou:


madig:
Hi Irene! Hold on, this seems to decompose a lot of components? Components are fine, but transformations other than moving them around won't variate. I.e. If you x-flip a component in the default master and y-flip it elsewhere, that won't work.

irenevlachou:
Hi Nikolaus,

you are right, I decomposed much more that I needed. I uploaded a new version, components now moving around, no flipping etc

madig:
There's still a lot of decompos. I thought only `dasia`, `dasia.case`, `lowernumeral-greek` need to be decomposed or something because they have different transformations across sources? 🤔 Or did you find more?

irenevlachou:
It was more than dasia the numeral sign. Caria was flipped as well and I though it was risky to keep two components side by side and then transformed again for the .case ones. I think it's better now and I fixed also some loose ones i.e oxia.sc and psili.sc that where non component glyphs for some reason.

madig:
This unfortunately seems to have broken outline compatibility, at least in the uprights (https://github.com/googlefonts/googlesans/pull/141/checks?check_run_id=1831171053#step:7:17):

> fontmake: Error: In 'GoogleSans/GoogleSans.designspace': Interpolating instance 'Medium' failed.: Failed to generate instance of glyph 'oxia': list index out of range. (Note: the most common cause for an error here is that the glyph outlines are not point-for-point compatible or have the same starting point or are in the same order in all masters.)

By the way, cascading transformations are perfectly fine, as long as they are the same across all sources.

Also, what do you use to work on the UFOs? Glyphs?

irenevlachou:
I fixed the compatibility errors. Thanks for the tip. I'm working on Robofont

chrissimpkins:
CI fails appear to be due to a glyph order change.  Nothing to do on your end about this Irene.  We'll have a look at the changes and push updated expected order definition files this week.  Be in touch if anything else is needed.  Thanks!

madig:
I may have to cherry-pick the changes, as the full change list includes e.g. the deletion of `source/GoogleSans/GoogleSans-Bold.ufo/glyphs._18_/G_.ss06.glif`, which is a separate issue.

madig:
I distilled the PR down to the actually relevant changes. There are a couple of changes I'm not sure about, e.g. does `dialytikaperispomeni` need to be an outline? Basically we want as few outlines as possible to keep file size down.

chrissimpkins:
OK to ignore the CLA notices.  Sorry for the distraction.  It is a bot and the CLA does not apply to this project.  Nothing to do.

chrissimpkins:
I discussed this with Irene today and she confirmed that the current state of decomposed components in the source here is how we will need to address this.  This will need to be rebased on the grade axis additions and expanded to all of the new GRAD -50 and GRAD +200 masters.

chrissimpkins:
@madig it looks like all of your questions were resolved in this thread.  Do you have any outstanding questions for Irene or are we ok to import these changes in the Greek script?

madig:
What's her answer to https://github.com/googlefonts/googlesans/pull/141#issuecomment-775353351?

chrissimpkins:
@madig no rush on this. Can you clarify all of the questions that are packed into the following comment when you have a chance?

> There are a couple of changes I'm not sure about, e.g. does dialytikaperispomeni need to be an outline?

madig:
I'm wondering if the PR contains only changes that need to be made. E.g. in https://github.com/googlefonts/googlesans/pull/141/files#diff-dfc51b1b089baa398bb9ac345c5835284ff2e8a33705ce8445b0bc3e3a8478a6L7, two components are replaced by outlines. Can they not stay components?

irenevlachou:
I've just pushed my latest commit that fixes the composite issue https://github.com/googlefonts/googlesans/commit/348664ce8c33816fe91cd4d3d531d5de9bb0b2b4

madig:
Hi @irenevl! I'm looking at getting this in and found... you're working with sources that are ungraded :O I think you may need to redo everything but on up-to-date sources and also consider the GRAD masters? @chrissimpkins or am I missing something here?

irenevlachou:
> Hi @irenevl! I'm looking at getting this in and found... you're working with sources that are ungraded :O I think you may need to redo everything but on up-to-date sources and also consider the GRAD masters? @chrissimpkins or am I missing something here?

yes-yes I know, I'm almost done in the GRAD masters too :) thanks @madig for picking this up!

chrissimpkins:
Superseded by https://github.com/googlefonts/googlesans/pull/384

PR #147, opened 2021-02-04
Add test for frac feature

madig:
Seems to be based on http://opentypecookbook.com/common-techniques/#method-2-contextual, so I added some basic tests to exercise things mentioned there.

Random find: there's no feature code to turn any fractions into precomposed ones like `seveneighths`.

### DIFF
--- a/qa/shaping/frac.json
+++ b/qa/shaping/frac.json
@@ -1,0 +1,189 @@
+{
+  "input": {
+    "text": [
+      "1/2 1/3 2/3 1/4 3/4 1/8 3/8 5/8 7/8",
+      "1234567890/1234567890123",
+      "11111111111/2222222222222",
+      "0/0/0 00/000/00000 0//0",
+      "2 1/2 222222222222 1/2"
+    ],
+    "features": {
+      "frac": true
+    },
+    "comparison_mode": "glyphstream"
+  },
+  "output": {
+    "GoogleSans-Italic[opsz,wght].ttf": {
+      "opsz=18.0,wght=400.0": [
+        "one.numerator|fraction|two.denominator|space|one.numerator|fraction|three.denominator|space|two.numerator|fraction|three.denominator|space|one.numerator|fraction|four.denominator|space|three.numerator|fraction|four.denominator|space|one.numerator|fraction|eight.denominator|space|three.numerator|fraction|eight.denominator|space|five.numerator|fraction|eight.denominator|space|seven.numerator|fraction|eight.denominator",
+        "one.numerator|two.numerator|three.numerator|four.numerator|five.numerator|six.numerator|seven.numerator|eight.numerator|nine.numerator|zero.numerator|fraction|one.denominator|two.denominator|three.denominator|four.denominator|five.denominator|six.denominator|seven.denominator|eight.denominator|nine.denominator|zero.denominator|one.denominator|two.denominator|three.denominator",
+        "one|one.numerator|one.numerator|one.numerator|one.numerator|one.numerator|one.numerator|one.numerator|one.numerator|one.numerator|one.numerator|fraction|two.denominator|two.denominator|two.denominator|two.denominator|two.denominator|two.denominator|two.denominator|two.denominator|two.denominator|two.denominator|two.denominator|two.denominator|two.denominator",
+        "zero|slash|zero|slash|zero|space|zero|zero|slash|zero|zero|zero|slash|zero|zero|zero|zero|zero|space|zero|slash|slash|zero",
+        "two|uni2009|one.numerator|fraction|two.denominator|space|two|two|two|two|two|two|two|two|two|two|two|two|uni2009|one.numerator|fraction|two.denominator"
+      ],
+      "opsz=18.0,wght=500.0": [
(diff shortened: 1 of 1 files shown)


chrissimpkins:
Thoughts about storing the configuration to gen these input text run and execution config values in TOML?  The input text run list is then more easily editable by a human and we pick up the ability to add comment annotations to the strings.

The shaping input data config for the file here would be:

#### frac.toml

```toml
[input.features]
frac = true

[input]
script = ""
language = ""
direction = ""
comparison_mode = "glyphstream"
text = [ 
    "1/2 1/3 2/3 1/4 3/4 1/8 3/8 5/8 7/8",   # common fractions
     "1234567890/1234567890123",   # default numerals
     "11111111111/2222222222222",   # big numerator/denominator
     "0/0/0 00/000/00000 0//0",  # fraction-like
     "2 1/2 222222222222 1/2",  # mixed fractions
]
```


```python
import json

import toml

print(json.dumps(toml.loads([TOML_STRING]), indent=2))
```

yields your JSON format ready to have the expected shaping data output dumped with https://github.com/googlefonts/googlesans/blob/master/qa/update_shaping_test_data.py and the appropriate set of static / variable font files.

```json
{
  "input": {
    "features": {
      "frac": true
    },
    "script": "",
    "language": "",
    "direction": "",
    "comparison_mode": "glyphstream",
    "text": [
      "1/2 1/3 2/3 1/4 3/4 1/8 3/8 5/8 7/8",
      "1234567890/1234567890123",
      "11111111111/2222222222222",
      "0/0/0 00/000/00000 0//0",
      "2 1/2 222222222222 1/2",
    ]
  }
}
```

The argument against moving the entire serialization approach over to a human readable/editable format that supports comments is that the output data needs to be embedded in the expected definition files and it looks like these files are going to be big and ugly. The five input text runs here yield almost 200 lines in the expected output values file.  It may be best not to hand edit these files.

We could go even simpler and just use newline delimited input text files with a different text run on each line along with a simple read filter for empty and comment lines.  We lose the ability to define all of the other required data for the shaping regression testing with that approach.

chrissimpkins:
Are you defining the expected output with the latest release builds, new builds through the remote GH Action approach, or is this coming from a local build with the requirements.txt defined build dependencies at the commit added here?

LGTM.  https://github.com/googlefonts/googlesans/pull/147#issuecomment-773689727 is commentary for possible future work.  I'm going to begin adding input strings to gen expected output files and am thinking through how it will be easiest to approach it.  I can create a local solution for now until we decide.

> Random find: there's no feature code to turn any fractions into precomposed ones like seveneighths.

As we come across bugs in production source let's open new issues and add the `bug` label.  Nothing to do about this one right now.  This is great that we are identifying these things as we turn over rocks.  Thanks for the detailed investigation Niko.

chrissimpkins:
> LGTM

Check that.  Is the new FB test running?

The current [master branch commit run shows two check-fea.py tests passing and eight skipped on the VF builds](https://github.com/googlefonts/googlesans/runs/1831267838#step:10:115) and the [latest commit here shows two tests passing and eight skipped on the VF builds](https://github.com/googlefonts/googlesans/pull/147/checks?check_run_id=1832074719#step:10:115).  

It looks like it may be skipped based on [an old commit that shows two passing and six skipped](https://github.com/googlefonts/googlesans/runs/1723545118#step:10:113) (not eight skipped as in the latest runs since the shaping test suite was added).  The two passes in the past have been the individual Roman and Italic feature tag checks.  Master branch should have two more skipped right now because there are no expected JSON files, one skip per vf file that is run through FB.  The commit in this PR should have two more PASS, not SKIP. (i.e. 4 PASS, 6 SKIP) = Roman and Italic VF through the new shaping regression check.  I think?

madig:
> Thoughts about storing the configuration to gen these input text run and execution config values in TOML? 

I made a separate issue to discuss improvements at https://github.com/googlefonts/googlesans/issues/149, please copy the comment there. I think TOML would be nice, but I need more tests to see if it would fit. Same for separate in- and output files. We're in the creative chaos phase right now!

> Are you defining the expected output with the latest release builds, new builds through the remote GH Action approach, or is this coming from a local build with the requirements.txt defined build dependencies at the commit added here?

That output was made with builds from master and its requirements. Incidentally, the test passes on build-armenian :)

> The current [...] skipped [...]

😨 will have a look.

madig:
So, running FB locally correctly picks up the frac test. I made a bug where the regression check will always _also_ return a SKIP:

```
 >> com.google.fonts/check/googlesans/features/regression
    But does it shape?
    with build/GoogleSans/static/expert/GoogleSans-Regular.ttf

    * PASS: qa/shaping/frac.json: No regression detected
    * SKIP: No test files found.
```

I fixed this in https://github.com/googlefonts/googlesans/pull/129/commits/da465da0e0464147ff89176350764ff0379b09e8. This is why I want to pull infra updates into master but am comfortable leaving the actual script imports sitting in their PR until the bed is made :)

chrissimpkins:
> So, running FB locally correctly picks up the frac test. I made a bug where the regression check will always _also_ return a SKIP:
> 
> ```
>  >> com.google.fonts/check/googlesans/features/regression
>     But does it shape?
>     with build/GoogleSans/static/expert/GoogleSans-Regular.ttf
> 
>     * PASS: qa/shaping/frac.json: No regression detected
>     * SKIP: No test files found.
> ```
> 
> I fixed this in [da465da](https://github.com/googlefonts/googlesans/commit/da465da0e0464147ff89176350764ff0379b09e8). This is why I want to pull infra updates into master but am comfortable leaving the actual script imports sitting in their PR until the bed is made :)

Confirmed

PR #155, opened 2021-02-05
Remove rogue Greek groups in feature files

### LINKED ISSUE #130: Features.fea: Greek groups redefined?
The Greek groups `@grk_UPPERCASE`, `@grk_lowercase` and `@grk_smallcaps` are defined twice, with the smaller definition winning out. What's the story here?

madig:
Final font files are bit-identical before and after, as determined by comparing TTX dumps.

Closes https://github.com/googlefonts/googlesans/issues/130.


madig:
Side-note: when trying to see if a change has an effect, set the `SOURCE_DATE_EPOCH` env var (set to some UNIX timestamp, output of `date +%s`) before `make` to get rid of useless timestamp and thereby checksum diffs in dumps.

Issue #149, opened 2021-02-04
Text shaping tests


madig:
Grab bag of thoughts I have on text shaping regression testing.

1. The tests do not help with conditional/bracket glyph testing in statics because statics have their outlines swapped instead of the actual glyph. VFs will show a different glyph name, but statics won't. Visual inspection is necessary.
2. If the output of a run is expected to be identical for all or most fonts, it would be nice to be able to specify this instead of having to look at pages full of identical lines.
3. Maybe the test files should have some more hierarchy and/or parameterization, so one can e.g. test different text runs with different features and/or languages in the same file. Think a `locl_latn` test document where you have the same text block but apply a diffent combination of languages (or even saying that a list of languages should behave the same).

Future potential improvements:

1. A GUI or repurposing of FontGoggles so designers can arrange a "worksheet" of fonts, a bunch of text blocks with feature, script and language toggles and the expected outcome, graphical and hb-shape-like. These worksheets can then be handed in along with the design work.

chrissimpkins:
Related #162 #163

chrissimpkins:
> Maybe the test files should have some more hierarchy and/or parameterization, so one can e.g. test different text runs with different features and/or languages in the same file. Think a locl_latn test document where you have the same text block but apply a diffent combination of languages (or even saying that a list of languages should behave the same).

This is the approach that I took in the PR's that I opened today (tracking in #161) when the behavior is known to differ across langs.  This is an issue for LTR vs. RTL writing systems too.  They will need separate files because of the input requirements for HB if we plan to make these direction definitions explicit rather than relying on HB.

MariannaPaszkowska:
Since we now use the Fontbakery checks, let's suggest those improvements upstream, if still applicable.

PR #158, opened 2021-02-05
Reduce diffs between features files

madig:
This will make maintenance much easier going forward.


madig:
Merging the feature files is probably going to throw around lookups, because I have to take out the italic locls and ss03 from the middle. Argh.

chrissimpkins:
Cyrillic `locl` regression tests in #164.  Waiting on your Latin `locl` regression tests.
`ss03` regression tests in https://github.com/googlefonts/googlesans/pull/174

Ready to merge if you are ok with the PR's.

.editorconfig
```diff
@@ -0,0 +1,16 @@
+# EditorConfig is awesome: https://EditorConfig.org
```
chrissimpkins: Let's split this file out into a different PR.  Not necessary here and I'd like to understand how this ~~conflicts~~ works with other formatters.  We can maintain formatting in the JSON files by only using the Python script to write them rather than editing them directly.  I had to avoid VS Code while working on the regression PR's this weekend because I have prettier formatting on by default.  It doesn't use the formatting that we are using here by default.  Let me know what you think about the TOML -> JSON approach in the open regression testing PRs.  The only thing that we would need to add is support for the git commit SHA writes and then we get the 2 space indent formatting defined in your Python script on all files without opening them in an editor.

qa/shaping/locl_latn_cat.json
```diff
@@ -0,0 +1,113 @@
+{
+  "input": {
+    "text": [
+      "L·L l·l",
+      "ŞŢşţ"
+    ],
+    "features": {},
+    "language": "ca",
```
chrissimpkins, line 8: Need three char tags or OK shorter so long as unique?

qa/shaping/locl_latn_mol.json
```diff
@@ -0,0 +1,113 @@
+{
+  "input": {
+    "text": [
+      "L·L l·l",
+      "ŞŢşţ"
+    ],
+    "features": {},
+    "language": "mo",
```
chrissimpkins, line 8: 2 vs. 3 char lang tags issue again

qa/shaping/locl_latn_rom.json
```diff
@@ -0,0 +1,113 @@
+{
+  "input": {
+    "text": [
+      "L·L l·l",
+      "ŞŢşţ"
+    ],
+    "features": {},
+    "language": "ro",
```
chrissimpkins, line 8: 2 vs. 3 char lang tags

source/GoogleSans/family_italics.fea
```diff
@@ -321,17 +321,17 @@ feature aalt {
     sub guilsinglleft.cap from [guilsinglleft.cap guilsinglleft];
     sub guilsinglright from [guilsinglright guilsinglright.cap];
     sub guilsinglright.cap from [guilsinglright.cap guilsinglright];
-	sub hookabovecomb from [hookabovecomb hookabovecomb.cap];
+    sub hookabovecomb from [hookabovecomb hookabovecomb.cap];
```
chrissimpkins, line 324: Does a fea linter exist?  We can document fea formatting style in our STYLE.md guidance.

source/GoogleSans/family_italics.fea
```diff
@@ -2126,26 +2125,26 @@ sub [space hyphen.cap endash.cap emdash.cap] Kappa' Alpha' Iota' [space hyphen.c
 
 sub [space hyphen softhyphen endash emdash underscore] kappa.sc' alpha.sc' iota.sc' [space hyphen softhyphen endash emdash underscore] by kaiSymbol.sc ;
 
-# =============
+# ===============
```
chrissimpkins, line 2128: These comment blocks are my doing.  Feel free to modify the format or remove, whatever allows us to understand how things fit together.

qa/shaping/locl_latn_cat.json
madig: I beleive HarfBuzz wants BCP 47 tags: https://www.iana.org/assignments/language-subtag-registry/language-subtag-registry

source/GoogleSans/family_italics.fea
madig: One can roundtrip files through feaLib, but it needs additional formatting work. I postponed this for later.

```python
import argparse
from pathlib import Path

import fontTools.feaLib.parser

parser = argparse.ArgumentParser()
parser.add_argument("feature_file", nargs="+", type=Path)
parsed_args = parser.parse_args()

for path in parsed_args.feature_file:
    fea = fontTools.feaLib.parser.Parser(path, followIncludes=False).parse().asFea()
    with open(path, "w") as f:
        f.write(fea)
```

source/GoogleSans/family_italics.fea
madig: That's ok, I just wanted to reduce diffs fore merging.

madig:
@chrissimpkins Let's rescope this to just reduce textual diffs and merge it before actually merging the feature files. I can get back to that after we have all the tests. 

This will now only change one thing, the order of `locl` lookups in uprights because it moved the CAT lookup around. Test is attached.

.editorconfig
madig: Removed for now.

madig:
@chrissimpkins this is ready to go.

moyogo:
Should there be a way to group font outputs together?
For substitution features the output is the same for all instances in most cases.

moyogo:
For example all the instances in the VF and the static fonts have the same output locl_latn.json.
https://github.com/googlefonts/googlesans/blob/841ee16d8a912388edee8b126b39c3ea8ffb7f2d/qa/shaping/locl_latn.json#L11-L19

madig:
Yes, and I noted this as a desireable improvment in https://github.com/googlefonts/googlesans/issues/149.

chrissimpkins:
Great minds think alike

moyogo:
Ah, sorry I missed that!

Issue #161, opened 2021-02-07
Add harfbuzz shaping regression tests for Latin, Greek, Cyrillic, Hebrew production scripts


chrissimpkins:
## High priority

- [x] calt (https://github.com/googlefonts/googlesans/pull/165)
- [x] case (https://github.com/googlefonts/googlesans/pull/166)
- [x] ccmp (Latin, Hebrew https://github.com/googlefonts/googlesans/pull/179)
- [x] dnom (https://github.com/googlefonts/googlesans/pull/168)
- [x] frac (https://github.com/googlefonts/googlesans/pull/147)
- [x] jalt (https://github.com/googlefonts/googlesans/pull/169)
- [x] kern (#180)
- [x] liga (https://github.com/googlefonts/googlesans/pull/170)
- [x] locl (Cyrillic https://github.com/googlefonts/googlesans/pull/164, Latin #189 )
- [x] numr (https://github.com/googlefonts/googlesans/pull/173)
- [x] ordn (https://github.com/googlefonts/googlesans/pull/172)
- [x] pnum (https://github.com/googlefonts/googlesans/pull/171)
- [x] sinf (https://github.com/googlefonts/googlesans/pull/173)
- [x] subs (https://github.com/googlefonts/googlesans/pull/173)
- [x] sups (https://github.com/googlefonts/googlesans/pull/173)
- [x] tnum (https://github.com/googlefonts/googlesans/pull/171)

## Medium priority

- [x] c2sc (#198)
- [x] dlig (https://github.com/googlefonts/googlesans/pull/167)
- [x] smcp (#190)
- [ ] liga + smcp strings (see https://github.com/googlefonts/googlesans/issues/69#issuecomment-784711656)

## Low priority

- [ ] lnum
- [x] ss01 (#205)
- [x] ss02 (#205)
- [x] ss03 (https://github.com/googlefonts/googlesans/pull/174)
- [x] ss04 (#190)
- [x] ss05 (https://github.com/googlefonts/googlesans/pull/183)
- [x] ss06 (https://github.com/googlefonts/googlesans/pull/184)
- [ ] ss07

## Will not test

- [x] aalt (to be removed in v4)
- [x] ss08 (to be removed in v4)

---

### TODO:

- [x] discuss Hebrew regression testing with Daniel Grumer
- [x] discuss Greek regression testing with Irene Vlachou
- [x] discuss Cyrillic regression testing with Alexei Vanyashin

chrissimpkins:
[...]

PR #165, opened 2021-02-07
Add harfbuzz shaping regression tests: calt

chrissimpkins:
This PR adds harfbuzz regression testing definition files for the `calt` feature.

The data have not been reviewed in detail to confirm that bugs do not exist in the production fonts. This intent is to add the production font data for regression testing and future bug hunt support.

### Issues

- I wonder if we want to *remove* support for replacement of the tabular colon with the colon.time (a figure centered design intended for NN:NN times), or better yet make a figure height centered design with the same tabular width.  If you are setting values with tabular figures, you likely want a tabular width colon too?

### Notes

- etatonos.sc sub is not currently included. We will need a separate definition file with SS06 and smcp set to true.  I haven't prioritized any of the smcp / stylistic set testing at this stage.

Tracking: https://github.com/googlefonts/googlesans/issues/161

### DIFF
--- a/qa/shaping/calt.json
+++ b/qa/shaping/calt.json
@@ -1,0 +1,295 @@
+{
+  "meta": {
+    "build_commit": "617d313bb59de7969f34439975febc98f7378403"
+  },
+  "input": {
+    "features": {
+      "calt": true
+    },
+    "script": "DFLT",
+    "language": "dflt",
+    "direction": "ltr",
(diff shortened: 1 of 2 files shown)


chrissimpkins:
Opened https://github.com/googlefonts/googlesans/issues/177 for the issue in OP.  smcp regression testing is being tracked in #161.  OK to merge.

PR #169, opened 2021-02-07
Add harfbuzz shaping regression tests: jalt

chrissimpkins:
This PR adds harfbuzz shaping regression testing definition files for the `jalt` feature.

The data have not been reviewed in detail to confirm that bugs do not exist in the production fonts. This intent is to add the production font data for regression testing and future bug hunt support.

Tracking: #161


chrissimpkins:
@danielgrumer I'd like to discuss this one with you so that I understand what happens with these forms.  We may want to add more text runs to this PR.

chrissimpkins:
Discussed with Daniel today.  Gtg here.

PR #181, opened 2021-02-09
Add language decls for cyrl BSH, CHU, SRB

### LINKED ISSUE #176: smcp and c2sc broken for BSH, CHU and SRB
See https://github.com/googlefonts/googlesans/pull/164.

madig:
Closes https://github.com/googlefonts/googlesans/issues/176.


source/GoogleSans/family_uprights.fea
```diff
@@ -9,6 +9,9 @@ languagesystem latn MOL;
 languagesystem latn CAT;
 languagesystem latn SRB;
 languagesystem cyrl BGR;
+languagesystem cyrl BSH;
```
chrissimpkins, line 12: Add MKD too?

chrissimpkins (changes requested):

chrissimpkins:
It looks like the bug may be in all features without explicit cyrl + one the lang tags here.  I tested tnum and numr in our prod files and they don't work in these locales.  Here is an action run with builds from this branch where they do:

https://github.com/googlefonts/googlesans/actions/runs/551384923

source/GoogleSans/family_uprights.fea
madig: There's no MKD use in uprights.

madig:
FWIW, I only needed to update the smcp files for this.

source/GoogleSans/family_uprights.fea
chrissimpkins: I'm trying to understand the root of this issue.  So if a locale is used anywhere explicitly in the feature source, you must add the languagesystem locale declaration for it to apply to anything outside of the block where it is defined (and also does not have an explicit locale definition that excludes it)?  Every locale falls under DFLT dflt, *unless* there is an explicit locale definition anywhere in the file in which case DFLT dflt does not encompass it?

chrissimpkins:
> FWIW, I only needed to update the smcp files for this.

Ready to merge?

madig:
I think so!

chrissimpkins:
FWIW, at some stage it might be useful to add lines for the localized glyphs only.  I have a tough time seeing them in the wall of Cyrillic glyphs... :)

chrissimpkins (approved):

madig:
Me too. We must prune, but first we must get everything in!

chrissimpkins:
changelog 8c03f101a932f07bffa918622982806c93adc24d

madig:
~~Wait. Now master MKD shaping is broken...~~

No, all good. My merge PR is broken probably.

PR #190, opened 2021-02-11
Add harfbuzz shaping regression tests: smcp

chrissimpkins:
This PR adds harfbuzz shaping regression tests for the smcp feature.


chrissimpkins:
@madig 57e3b5e753f023220a2d03108d72a0054013b252 includes the ss04 `g` and `y` with bugs. Known issue that we are addressing in #139

chrissimpkins:
It looks like there are a number of other .BRACKET.18 glyphs without .sc extensions in the smcp set that we'll need to comb through. rvrn, ugh.

PR #134, opened 2021-01-21
Hebrew scaled dagesh

moyogo:
The dagesh dot used in some glyphs has different scaling transformation than the default master in some masters.
This is problematic in variable fonts, since they can only use the scaling transformation from the default master.

- [x] Fix the variable fonts dagesh dot size in pedagesh-hb, pedagesh-hb.BRACKET.18, finalpedagesh-hb, finalpedagesh-hb.BRACKET.18, shindagedh-hb, shindagedhsintdot-hb, shindagedhshindot-hb
  - [x] Add a scaled dagesh-hb.alt with the transformation used in composite glyphs
  - [x] Update composite glyphs
  - [x] Add dagesh-hb.alt to glyhpsetdef

Split from #118 as this doesn't require to update the toolchain requirements.

PR #193, opened 2021-02-11
Regression test for latn ccmp

madig:
https://github.com/googlefonts/googlesans/issues/161#issuecomment-777659783


madig:
All data from v3.003.

Observations:
1. There's no soft-dot ccmp rule for jdotless.alt -- but those are going to go away anyway?
2. Mark-stacking works, but .cap variants aren't used for caps. Small-caps seem to work.
3. The small-caps variants for idotless (i.sc) and jdotless (j.sc) aren't swapped in -- or is that expected? What is the ccmp lookup for soft-dots supposed to do for smcp? @moyogo

moyogo:
> 1. There's no soft-dot ccmp rule for jdotless.alt -- but those are going to go away anyway?

Oh, are we removing jdotless.alt?

> 2. Mark-stacking works, but .cap variants aren't used for caps. Small-caps seem to work.

Good point. The `sub brevecomb.sc acutecomb.sc by brevecomb_acutecomb.sc;`, etc. are not necessary as there is no .sc glyphs at that point (ccmp lookups are before c2sc and smcp lookups).

In theory the @Markscomb and @MarkscombCase should have contained the proper combining marks and their .cap variants, and the rule that converts them when preceded by an uppercase or a .cap variant should be kept.
That way `sub brevecomb.cap acutecomb.cap by brevecomb_acutecomb.cap;`, etc. would be useful.

>  3. The small-caps variants for idotless (i.sc) and jdotless (j.sc) aren't swapped in -- or is that expected? What is the ccmp lookup for soft-dots supposed to do for smcp? @moyogo

They should be swapped in the smcp with `sub idotless by i.sc; sub jdotless by j.sc;` like the dotted ones.

madig:
Regarding point 2 and 3, is that something you'll bring in in your Latin expansion?

moyogo:
> Regarding point 2 and 3, is that something you'll bring in in your Latin expansion?

Yes. I'll make sure that's all in.

madig:
Re point 1, Chris said that .alt glyphs may be on the chopping block after he confirms with Crossland and users.

chrissimpkins:
> Re point 1, Chris said that .alt glyphs may be on the chopping block after he confirms with Crossland and users.

We plan to eliminate stylistic alternates.   Creative oversight on the project decided that they do not want to allow internal design teams to decide the combination of alternate forms that represent the brand.  The intent is to fix branding with a single set of styles.  They haven't gone away to date because they've been in production for a couple of years and we don't have a good way to understand who is using them now, nor a way to signal that they are going to be eliminated.  

The Fonts API has always stripped `aalt` feature support out so there are no web font use cases that have ever had access to those alternates.   

When we eventually address these .alt glyphs we need to note that there are those like the new dagesh that Denis just merged in #134 that *do* need to stay.  This won't be a broad scale removal of all *.alt.  I don't know what else in that set addresses issues that are not purely stylistic alternate forms.

chrissimpkins:
> There's no soft-dot ccmp rule for jdotless.alt -- but those are going to go away anyway?

I can confirm that all stylistic alternates that are supported through the `aalt` feature will be removed as of our upcoming v4 release.  

This gtg?

madig:
I think so. Denis can expand on it in his extension.

PR #196, opened 2021-02-12
Trim cyrl smcp tests

madig:


chrissimpkins:
Ty!

I think that we have nothing in master at this stage that limits us from using 2d2b70d6710f1624e645c2ef5abbe324880df5c8 as the base commit for the regression tests here.  Let's standardize this to the [last release build ](https://github.com/googlefonts/googlesans/releases)for anything that is already in production going forward.

madig:
You mean build commit 2d2b70d and update all JSON files with the behavior of that commit? Or just use it for new tests?

chrissimpkins:
No need to build anything to address my request.  The builds are in the release as zip archives.  The commit that we used for these production pushes is listed in the release too.   I think that we regression test against existing behavior in production with those builds that were pushed to production.  This is for current production behavior only.

For all new behavior that comes in  (new script merges, expansion/reduction of features, bug fixes), we will need to build fonts and update the regression tests based on the code state at that time.  

This PR doesn't follow that approach.  Nothing to do.  Let's use the above approach going forward.

PR #198, opened 2021-02-12
Test smcp _and_ c2sc

madig:
I copy-pasted the lowercase string and uppercased it, removing some Greek stuff that has no uppercase.


madig:
@chrissimpkins anything missing here from your perspective?

chrissimpkins:
Thanks Niko!

PR #203, opened 2021-02-23
Add hb shaping regression testing definition file toml2json conversion support

chrissimpkins:
Adds a new toml dependency in the `requirements-dev.txt` file and a toml -> json format conversion in the `update_shaping_test_data.py` script.

The script now takes a toml file argument instead of a json file argument.  The toml is converted to json and then the hb shaping data write takes place on the json definition file as before.  The toml file path stem is used for the json path.  

We still need to write the git commit SHA manually at this stage.

### DIFF
--- a/qa/update_shaping_test_data.py
+++ b/qa/update_shaping_test_data.py
@@ -156,7 +156,7 @@ if __name__ == "__main__":
     if shaping_file_toml.exists():
         toml2json(shaping_file_toml)
     else:
-        sys.stderr.write(f"{shaping_file_toml} does not appear to be a valid file!")
+        sys.stderr.write(f"{shaping_file_toml} does not appear to be a valid file!\n")
         sys.exit(1)
 
     # open json file that was dumped above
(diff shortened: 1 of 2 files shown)


qa/update_shaping_test_data.py
```diff
@@ -124,13 +124,23 @@ def shape_texts(
...
 
 
+def toml2json(tomlpath: Path) -> None:
+    with tomlpath.open() as tf:
+        json_filepath = get_json_filepath(tomlpath)
+        json_filepath.write_text(json.dumps(toml.load(tf), indent=2, ensure_ascii=False))
+
+
+def get_json_filepath(tomlpath: Path) -> Path:
+    return Path("shaping", f"{tomlpath.stem}.json")
```
madig:
```suggestion
    return Path("shaping", tomlpath.with_suffix(".json").name)
```

chrissimpkins: Mine was shorter! ;)

Issue #214, opened 2021-02-24
Transition Latin glyph set optical size rvrn substitution support to compatible opsz interpolation designs + avar table


chrissimpkins:
We are running into a number of problems with optical size axis rvrn support and the approach to address these issues involves glyph duplication with a file size increase.  Our goal has been to "eliminate" opsz interpolation over the min opsz=17 master : max opsz=18 master range.  Ideally, there should be no visible interpolation between opsz 17 and opsz 18, including glyphs that are compatible forms now.

David Berlow worked on compatible versions of the Latin opsz 17 -> 18 rvrn designs and we have an approach to implement this with an avar table that maintains the optical size axis value dependent glyph substitution behavior (https://github.com/googlefonts/googlesans/issues/201).  We will try to establish the abrupt rvrn style substitution behavior as opsz is defined at 18+ across as many of these glyphs as possible.

We will maintain all rvrn substitutions that are included in the Hebrew glyph set for now.

This transition will occur as of the Google Sans v4 release.

chrissimpkins:
@danielgrumer Daniel, we won't be able to fit it into this release but I would appreciate it if you would be willing to have a look at all of the Hebrew rvrn (Glyphs bracket layer) designs and let us know if it will be possible to transition to compatible forms so that we can eliminate the rvrn substitutions and use the same approach in a future point release.  With the avar table, you will not see the interpolated intermediate forms.

chrissimpkins:
Current rvrn definitions:

https://github.com/googlefonts/googlesans/blob/67508ad446a9105a1f25b02d3f4dbe32be138d10/source/GoogleSans/GoogleSans.designspace#L12-L70

chrissimpkins:
rvrn substitutions that will need to remain after the merge of the grade axis expansion work from David Berlow:

- finalpe-hb
- finalpedagesh-hb
- pe-hb
- pedagesh-hb
- perafe-hb
- samekh-hb
- samekhdagesh-hb

`G`, `y`, `semicolon`, `comma`, and `commaccent`, etc from Latin set have compatibilization work in progress.  We are attempting to transition away from rvrn altogether in the Latin set.

chrissimpkins:
@madig these forms will not be completely sorted by tomorrow.  I suggest that we begin the merge and work in parallel with David while he works through the designs to fully address this issue.  The issues at this stage are the following:

[...]

Issue #211, opened 2021-02-24
Remove stylistic alternates and aalt feature code


chrissimpkins:
Project creative directors asked that we remove support for alternate glyph forms so that internal design teams do not have the opportunity to select from the alternate forms to define branding in their own product areas.

This change will occur as of the v4 release of Google Sans.  The source definitions will remain but the glyphs and feature code support will not be compiled into the production fonts.  As part of this transition we will eliminate the "default" and "expert" build targets and settle on a single default build that is compiled into the top level of the build/GoogleSans directory.

We are aware that some forms are **not** stylistic alternates (e.g., Denis Jacquerye's dagesh-hb.alt) and will need to remain in the build.  Full review of the required *.alt glyphs to follow.

chrissimpkins:
@danielgrumer The `aalt` feature is subset out in the Fonts API so this is not available to anyone using the web fonts.  We are going to remove all other `aalt` feature support in the v4 release.  Can the justification alternates be removed from the `aalt` feature block so that we can get rid of `aalt` altogether?

chrissimpkins:
I reviewed all of the *.alt alternate forms.  The following alternate style glyphs can be set to not export / not compile into fonts:

```python
[...]

PR #216, opened 2021-02-25
Import Berlow sources

madig:
1. Add new grading UFOs and overwrite all existing glyphs directories, groups and kerning with Berlow sources, verbatim.
2. Remove `aalt` feature.
3. Set `.alt` glyphs to not export.

TODO:
- [ ] Unfudge production scripts
    - [ ] Adjust partial instancer/Makefiles to produce graded instances?
    - [ ] Get rid of default/export distinction?
- [x] Reinstate `rvrn` where necessary
- [ ] Recheck glyphOrder and postscriptNames for the final fonts to be imported
- [ ] Check final fonts to be imported for leftover glyphs
- [ ] Update the shaping tests
- [ ] Check fontinfo metadata, especially the interpolatable part
- [ ] ?


chrissimpkins:
> Adjust partial instancer/Makefiles to produce graded instances?

Unnecessary at the moment.  High priority build format is Roman and Italic variable fonts with all axes.  Not necessary to create new named instances for grades in VF's either.  Let's keep the same named instances that we had as of v3.003.

> Get rid of default/export distinction?

Yes please.  Let's eliminate the `expert` and `default` sub-dir structure. We'll only build one default build target and it isn't necessary to place it in a sub-dir.  We'll keep the static and variable dirs for the separate build formats.  All defined in Makefile variables currently

> Set .alt glyphs to not export

Copy / paste Python list https://github.com/googlefonts/googlesans/issues/211#issuecomment-785516852

> rvrn

Changes in progress and not stable downstream yet

> ?

I'm with you on the `?`  Let's venture into the unknown :)

madig:
K, so I will leave `rvrn` out until I hear otherwise.

chrissimpkins:
Can we agree on the following build artifact paths?  I'll watch for where you end up today and can dig in on GH Actions and build scripting file path updates

### Variable

- `build/GoogleSans/variable/GoogleSans[GRAD,opsz,wght].ttf`
- `build/GoogleSans/variable/GoogleSans-Italic[GRAD,opsz,wght].ttf`

### Static instances

- `build/GoogleSans/static/GoogleSans-Regular.ttf`
- `build/GoogleSans/static/GoogleSans-Italic.ttf`
- `build/GoogleSans/static/GoogleSans-Medium.ttf`
- `build/GoogleSans/static/GoogleSans-MediumItalic.ttf`
- `build/GoogleSans/static/GoogleSans-Bold.ttf`
- `build/GoogleSans/static/GoogleSans-BoldItalic.ttf`
- `build/GoogleSans/static/GoogleSansText-Regular.ttf`
- `build/GoogleSans/static/GoogleSansText-Italic.ttf`
- `build/GoogleSans/static/GoogleSansText-Medium.ttf`
- `build/GoogleSans/static/GoogleSansText-MediumItalic.ttf`
- `build/GoogleSans/static/GoogleSansText-Bold.ttf`
- `build/GoogleSans/static/GoogleSansText-BoldItalic.ttf`

madig:
Can implement tomorrow!

chrissimpkins:
> Can implement tomorrow!

Ty!  Wanted to make sure that you are ok with those paths.  I'll begin to look into the changes outside of the Makefiles

qa/check-fea.py
```diff
@@ -131,7 +128,6 @@
     "numr",
     "ordn",
     "pnum",
-    "rvrn",
```
chrissimpkins: I think that we will have to keep this to support Hebrew rvrn for now (https://github.com/googlefonts/googlesans/issues/214#issuecomment-785435774).  These won't be transitioned to compatible forms in time for the v4.000 release but we'll look into it.  David B is trying to eliminate it altogether in the Latin where we are facing bugs.  I think that more rvrn may be coming in with Armenian?  We may not get away from it altogether.  Let's see.

madig: Ah, right. I guess I'll leave in rvrn for non-LGC.

PR #232, opened 2021-02-26
Update GH Actions CI test paths to new v4 build artifacts, update release workflow with version string in font archive

### LINKED ISSUE #100: Improve build artifact ZIP
Version 3.003 (from https://github.com/googlefonts/googlesans/releases has a `GoogleSans.zip` which unpacks as `/build\ 2/GoogleSans`

I have 2 requests

- [ ] Filename can declare version, eg `GoogleSans-v3_003.zip`
- [ ] Path should not include the first `/build\ 2/GoogleSans` empty top level folders

chrissimpkins:
Updates the GitHub Action CI, test font upload, and release prep/push workflows and associated Python scripts with new build artifact paths that will be used as of GS v4.  

This also adds support for version strings in the GitHub release zip archives (#100).

Closes https://github.com/googlefonts/googlesans/issues/100

Requires #240 

cc @madig


madig:
@googlebot I consent.

madig:
In my commit, I removed any "default" variant. The fonts are now always the "expert" set. `aalt` can be removed from the sources directly for v4. It also removes instancing.

madig:
As I see it, the only point left for the subsetter is to remove unused glyphs from the statics (NULL funnily enough, all BRACKET glyphs). If `rvrn` were to be completely eliminated, we could take out the subsetter, too.

chrissimpkins:
> As I see it, the only point left for the subsetter is to remove unused glyphs from the statics (NULL funnily enough, all BRACKET glyphs). If `rvrn` were to be completely eliminated, we could take out the subsetter, too.

👍 

I was looking into the rvrn issue today.  It looks like we will be able to transition all of the Latin & punctuation (including combining mark forms that were derived from punctuation) rvrn over to compatible interpolated forms + avar.  There are a handful of Hebrew glyphs that will still require rvrn support.  So the corresponding *.BRACKET.18 glyphs will remain.  Thanks for having a look at this Nikolaus.  It sounds like we keep the subsetter.

chrissimpkins:
Rebased on the GH Actions config changes from #237 after the default branch change

madig:
Huh? GitHub took care of all master->main rebasings in othr projects :thinking:

chrissimpkins:
Not in GH Actions configs...

madig:
Oh, duh. I was thinking of just a branch rebasing.

chrissimpkins:
We made GHA changes in this PR and I didn't know if it would lead to a merge conflict.  I just updated the master -> now main branch with updated GHA config files with a rebase here.  Should be all set. I don't think that we touched GHA in any other branch/open PR.

PR #239, opened 2021-03-02
[fontbakery tests] Update expected feature tags for GSv4 builds

chrissimpkins:
removes ss08 and aalt feature tags from the expected list.  These features will be removed in GSv4

This update should fail CI testing until the updated GSv4 source is merged (#240)


chrissimpkins:
No longer required.  This was added in #240

PR #265, opened 2021-03-06
Eliminate grade on logo forms, Roman only

### LINKED ISSUE #249: Logo ligatures have GRAD weight change in Roman only
The logo form set has a weight change on the GRAD axis in Roman forms only.

@dberlow I seem to recall that you planned not to use grade on these forms?

logos

chrissimpkins:
We do not want grade on the logo at this time.  

Italic does not support grade. No change is required in the Italic glyph set. 

Fixes #249


chrissimpkins:
Confirmed that these changes to do not alter adv width across GRAD masters

chrissimpkins:
Compiled locally.  Fix confirmed.  LGTM

PR #275, opened 2021-03-10
Update all harfbuzz shaping regression definitions to GS v4.000 prod file commit build artifacts

chrissimpkins:
WIP

- Updates to commit e2766096416e641da5edff0ced5c1f8f9dcc0e5b
- Fixes file path order based on the definitions in our qa/update_all_shaping.sh script to avoid future diffs due to the order of file path args

--- 

### TODO

- [x] review diffs, these should only be file path order changes, not differences in shaping data
- [x] add toml files where missing
- [x] update json files with git commit SHA where missing (due to missing toml)


madig:
Ah yes, I think this should go in first before Armenian and Deva (after I rebase it) to cut down on test suite diff churn.

chrissimpkins:
Sg.  I will try to review and merge today.  Thanks Niko.

chrissimpkins:
Added missing frac.toml file and updated frac.json.

PR #290, opened 2021-04-12
Fix generate glyphs workflow

### LINKED ISSUE #188: "Generate Glyphs.app file" workflow is broken
https://github.com/googlefonts/googlesans/runs/1873002807?check_suite_focus=true#step:6:1

Doesn't do anything.

madig:
Closes https://github.com/googlefonts/googlesans/issues/188.

I can't remember why I used the staging directory, it makes more sense to generate fresh Glyphs files from the main source dir for merges into the main source.


chrissimpkins:
gtg?

madig:
I think so.

PR #289, opened 2021-04-12
Add Noto OS/2 table Unicode range bit flag checks to custom fontbakery profile

chrissimpkins:
This PR adds a slightly modified version of the [Noto OS/2 table UnicodeRange bit flag check vs. code points in a font](https://github.com/googlefonts/fontbakery/blob/58b65965449a1032fa6b77a7f1807d106717eab2/Lib/fontbakery/profiles/notofonts.py#L88-L119).  The modification makes bits that are set without any code point in range a FAIL instead of a WARN.   

A new script to address this in our build workflow will follow.  We should see a number of new WARN for unset bits with this new check.

PR #291, opened 2021-04-12
Add automated OS/2.ulUnicodeRange bit flag settings based on available code points to build workflow

chrissimpkins:
Automates OS/2 table ulUnicodeRange bit flag fields 1 - 4 settings. If at least one in range code point is available in the build artifact, the corresponding bit is set.  Clears bit if there are no code points in range.

Related: #289

### DIFF
--- a/scripts/set-unicoderange.py
+++ b/scripts/set-unicoderange.py
@@ -1,0 +1,407 @@
+#!/usr/bin/env python3
+
+#    Copyright 2021 Source Foundry Authors
+#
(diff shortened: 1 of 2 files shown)


chrissimpkins:
D/W Niko and Jany.  There is agreement that an aggressive approach to setting these bit flags is useful, primarily for support in MS Office apps.  We are now passing the new FB check for the bit flag settings.  Will merge and we can decide if we want to transition to the fonttools implementation in the future.

PR #299, opened 2021-05-14
Use pypy3

### LINKED ISSUE #298: Using pypy3 to speed up the CI
It is possible to use pypy3 on GitHub Actions (see https://github.com/fonttools/fonttools/blob/b77d520/.github/workflows/test.yml#L79-L91) and the final font binaries are bit-identical, but the speed gains vary (see caveats below).

| Target | CPython 3.8            | Pypy 3.7.4  |
| --------- | ------------------- | --------- |
| gs-static | 250s | 200s |
| gs-vf | 320s | 180s |

These numbers came about with the current library versions, which include booleanOperations (sloooow!) and ufo2ft and fonttools that don't have the latest optimizations. Source is `build-gurmukhi`. CPU is a Ryzen 5 3400G, run inside WSL2.

I will test again after we upgraded libraries and maybe optimized the build system somewhat (parallelization).

madig:
This is an experimental PR using Pypy instead of CPython, which should speed up CI: https://github.com/googlefonts/googlesans/issues/298

I recommend we don't merge this until we have landed the big script PRs, to not accidentally upset something.

Closes #298

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -16,8 +16,6 @@ booleanoperations==0.9.0
     #   ufo2ft
 cffsubr==0.2.7
     # via ufo2ft
-compreffor==0.5.0
-    # via ufo2ft
 cu2qu==1.6.7
     # via
     #   fontmake
@@ -26,7 +24,7 @@ fontmake==2.2.0
     # via -r requirements.in
 fontmath==0.6.0
     # via fontmake
-fonttools[lxml,ufo,unicode]==4.19.1
+fonttools[ufo,unicode]==4.19.1
     # via
     #   booleanoperations
     #   cffsubr
@@ -41,8 +39,6 @@ fs==2.4.11
     # via fonttools
 glyphslib==5.3.2
     # via fontmake
-lxml==4.6.3
-    # via fonttools
 pyclipper==1.2.0
     # via booleanoperations
 pytz==2020.1
@@ -58,3 +54,5 @@ ufolib2==0.8.0
 
 # The following packages are considered to be unsafe in a requirements file:
 # setuptools
+
+--no-binary=cu2qu
--- a/requirements.txt
+++ b/requirements.txt
@@ -56,3 +56,4 @@ ufolib2==0.8.0
 # setuptools
 
 --no-binary=cu2qu
+--no-deps
--- a/.github/workflows/lang-coverage.yml
+++ b/.github/workflows/lang-coverage.yml
@@ -25,7 +25,7 @@ jobs:
       - name: Check out ${{ steps.config.outputs.projectname }} source repository
         uses: actions/checkout@v2
       - name: Set up Python v${{ steps.config.outputs.py-version }} environment
-        uses: actions/setup-python@v1
+        uses: actions/setup-python@v2
         with:
           python-version: ${{ steps.config.outputs.py-version }}
       - name: Python build dependency cache lookup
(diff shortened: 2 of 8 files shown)


requirements.txt
```diff
@@ -58,3 +54,5 @@ ufolib2==0.8.0
 
 # The following packages are considered to be unsafe in a requirements file:
 # setuptools
+
+--no-binary=cu2qu
```
chrissimpkins, line 58: How does this impact the release workflow?  Should probably transition our release approach to use the same set of dependencies that we are using to test?

chrissimpkins: https://github.com/googlefonts/googlesans/blob/main/.github/workflows/release.yml
https://github.com/googlefonts/googlesans/blob/main/.github/workflows/upload-artifacts.yml

chrissimpkins:
CI seems to have crumped for some reason.  I launched all of the workflows again.

chrissimpkins:
CI continues to choke on this PR for some reason. I wonder if it is having difficulty pulling the pypy3 interpreter. Unclear what is going on.  I can't see a log.

https://github.com/marketplace/actions/setup-python#available-versions-of-pypy

requirements.txt
madig: This is a special instruction to pip to not use pre-built wheels for cu2qu, because we specifically want the Python-only code for JITting. It's still the same dep!

madig:
Yeah, I don't know why CI pukes here...

madig:
I think it was an outage. 

Also: https://github.com/googlefonts/googlesans/pull/299/checks?check_run_id=2599444298 something is prepending "v" to the Python version, making GHA fail.

chrissimpkins:
@madig I think that the prepended `v` that you are seeing is just in the title of the action step and is due to the string formatting in this line:

https://github.com/googlefonts/googlesans/blob/74fd958c7b779323d05419843cd2d88a8287c20d/.github/workflows/static-artifact-ci.yml#L27

This seems to be defining the appropriate pypy-3.7 interpreter.

chrissimpkins:
It looks like this is the problem with the `actions/setup-python` step:

```
Error: Version pypy-3.7 with arch x64 not found
```

chrissimpkins:
Found the problem.  We need to update to actions/setup-python@v2:

https://github.com/actions/setup-python/releases/tag/v2.2.0

chrissimpkins:
@googlebot I consent.

chrissimpkins:
> Found the problem. We need to update to actions/setup-python@v2:
> 
> https://github.com/actions/setup-python/releases/tag/v2.2.0

https://github.com/googlefonts/googlesans/pull/299/commits/038974b0f46374abb294f277b56d581e8fbe9631

madig:
I think I'm going to postpone the switch. To do it cleanly, I have to 

1. change fontmake's fixed dependency on lxml (no available wheel for pypy) at https://github.com/googlefonts/fontmake/pull/773 and also 
2. figure out how to make wheels for pathops https://github.com/fonttools/skia-pathops/pull/46

Both efforts are hindered by GHA just not running workflows for some reason???

I suggest we kick the can down the road until the big scripts are merged and we can do a general dependency update. There's some good optimizations in newer ufo2ft and fontTools that we're not using yet!

chrissimpkins:
> I think I'm going to postpone the switch. To do it cleanly, I have to
> 
> 1. change fontmake's fixed dependency on lxml (no available wheel for pypy) at [googlefonts/fontmake#773](https://github.com/googlefonts/fontmake/pull/773) and also
> 2. figure out how to make wheels for pathops [fonttools/skia-pathops#46](https://github.com/fonttools/skia-pathops/pull/46)
> 
> Both efforts are hindered by GHA just not running workflows for some reason???
> 
> I suggest we kick the can down the road until the big scripts are merged and we can do a general dependency update. There's some good optimizations in newer ufo2ft and fontTools that we're not using yet!

sgtm!

PR #300, opened 2021-05-14
Transition to pypy3 Python interpreter for CI testingTransition to pypy3 Python interpreter for CI testing

chrissimpkins:
Transitions from cPy 3.8 to pypy3.7 interpreter in the CI testing workflows


chrissimpkins:
Bah...

PR #301, opened 2021-05-14
Add individual OpenType table file size assessment to our static and VF CI pipeline

chrissimpkins:
Supports our upcoming file size optimization experimentation/testing


chrissimpkins:
@madig @belluzj This adds individual OT table sizes to the previous summary statistics that were available in the CI. This includes the GPOS table for the kerning/spacing experiments

PR #310, opened 2021-06-02
Georgian script import

chrissimpkins:
Supersedes #93 
Tracking #309 
Closes #309 

- Add Georgian script Roman and Italic sources to the `source/GoogleSans/GEORGIAN` directory (https://github.com/googlefonts/googlesans/commit/707be3fc72142bff4c009d8e795269745393b9ba)
- Add `source/GoogleSans/GEORGIAN/import-glyph-list.txt` with the glyphs that will be merged from Glyphs sources => production UFO sources (https://github.com/googlefonts/googlesans/commit/42e14525ba8727632fd80805e70de978fb492436)

New lang systems:

- `geor dflt`
- `geor KAT`

New feature support:

- case
- locl
- calt
- tnum (lari currency symbol)

### MERGE PREP TODO's:

- [x]  add toml shaping files with text runs recommended by Akaki

### POST-MERGE TODO:

- [x] `semicolon.loclGEO` was edited in the Glyphs sources that were pushed in https://github.com/googlefonts/googlesans/commit/707be3fc72142bff4c009d8e795269745393b9ba with (1) removal of all [18] bracket layers; (2) revise semicolon outlines to components of the default semicolon form.  I discussed this issue with Akaki and he will review, make edits to the sources following the initial merge so that we can add his localized semicolon forms *without* rvrn subs support. 
- [x] confirm tnum shaping output shows expected lari tabular form
- [x] confirm case shaping output shows expected caps forms
- [x] confirm calt shaping output shows expected punctutation v position adjustments vs. LC forms
- [x] confirm locl shaping output shows expected colon.loclGEO and semicolon.loclGEO forms
- [x] update expected glyph set / order definition files
- [x] remove commented out Georgian fea code (https://github.com/googlefonts/googlesans/pull/310#issuecomment-856910240) 
- [x] remove Georgian calt fea code (code added to #328 to preserve designer intent)
- [x] georgian.fea group namespacing, possibly add `geor_*` prefixes on these group names?
- [x] remove `punctuation_centered` fea group (list added to #328 to preserve designer intent)
- [x] remove `languagesystem geor HYE;` in `family_uprights.fea` and `family_italics.fea`
- [x] remove ` script geor; language HYE; lookup armn_locl_shift_guillemet;` in `family.fea`

### References

- designer intent calt code for Georgian script is documented in https://github.com/googlefonts/googlesans/pull/310#issuecomment-867914010

---

Akaki's original source files are in the archive below:

[...]

PR #319, opened 2021-06-17
Update dependencies

madig:
The only binary diff I see is an added 

```
          <PosLookupRecord index="0">
            <SequenceIndex value="0"/>
            <LookupListIndex value="3"/>
          </PosLookupRecord>
```

at the end of some `ChainContextPos` blocks. Not sure what that means.


chrissimpkins:
> The only binary diff I see is an added
> 
> ```
>           <PosLookupRecord index="0">
>             <SequenceIndex value="0"/>
>             <LookupListIndex value="3"/>
>           </PosLookupRecord>
> ```
> 
> at the end of some `ChainContextPos` blocks. Not sure what that means.

Sounds like good news.  This includes all updates required to support the GPOS optimization work?

chrissimpkins:
Hmm... shaping tests are failing though

madig:
Yeah, specifically, georgian_calt, where the punctuation after the Latin is set down unconditionally. Trying to replicate in the browser...

madig:
Weird. I don't see this in the browser (neither Firefox nor Chrome, HarfBuzz 2.8.1):

```
> hb-shape --text "abo- აბო-" OLD/GoogleSans\[GRAD,opsz,wght\].ttf
[a=0+531|b=1+602|o=2+596|hyphen=3+441|space=4+232|uni10D0=5+506|uni10D1=6+572|uni10DD=7+768|hyphen=8@0,-25+441]

> hb-shape --text "abo- აბო-" NEW/GoogleSans\[GRAD,opsz,wght\].ttf
[a=0+531|b=1+602|o=2+596|hyphen=3@0,-25+441|space=4+232|uni10D0=5+506|uni10D1=6+572|uni10DD=7+768|hyphen=8@0,-25+441]
```

chrissimpkins:
Possibly due to a Harfbuzz update in the uharfbuzz lib on our end?

madig:
Hm.. not sure. I mean there _is_ a binary diff and the same HarfBuzz produces different results for old and new 🤔 Maybe there's something else going on in browsers?

chrissimpkins:
We can get at harfbuzz shaped visuals outside of the browser with Just's [Font Goggles](https://fontgoggles.org/).  I believe that if you use different file paths for A and B they can be viewed simultaneously, even with the same family name.

madig:
Hm, yeah. I'll have a look next week.

madig:
It doesn't show in FontGoggles either, need to dig deeper.

madig:
Still don't understand this. So this calt lookup gained an additional PosLookupRecord, which IIUC applies lookup 3 to glyph in position 0, so here, to hyphen. It seems to do so backwards, which explains why it hits the hyphen in front of the Georgian characters I suppose?

image

It does not trigger if we put more Latin stuff in front:

```
> hb-shape --language kat --text "abo- abo- აბო-" build/GoogleSans/variable/GoogleSans\[GRAD,opsz,wght\].ttf
[a=0+531|b=1+602|o=2+596|hyphen=3+441|space=4+232|a=5+531|b=6+602|o=7+596|hyphen=8@0,-25+441|space=9+232|uni10D0=10+506|uni10D1=11+572|uni10DD=12+768|hyphen=13@0,-25+441]
```

Maybe something actually didn't work before but does so now?

image

@razmadzekoko what is the intended behavior here?

razmadzekoko:
Hi, 

Screenshot 2021-06-22 at 20 01 49
this code line is defined for shifting down some characters, which are located in the class @punctuation_centered so for example: if we have hyphen-minus, then comes space and then comes one Georgian letter like this: - ო the hyphen-minus should be shifted down with 25 points.

madig:
Hm... so in strings like `abo- აბო-`, do you actually expect both hyphens to be shifted downwards?

razmadzekoko:
> Hm... so in strings like `abo- აბო-`, do you actually expect both hyphens to be shifted downwards?

No, only at აბო-
Latin letters are just for comparing, so we can see that hyphens next to Georgian letter is really shifted down.
The hyphen after Latin letters should **not** be  shifted.

razmadzekoko:
> No, only at აბო-

and in other cases where hyphens and similar characters are before or after Georgian letters.
So it can be: 
აბო-
-აბო
ა-ბო
ა -ბო
ა- ბო
აბო,-
and other cases

madig:
Jany figured it out!!!!

We are looking at a bugfix. fontTools 4.23.0 contains a fix for https://github.com/fonttools/fonttools/issues/2293, which happens to affect the following code:

```fea
    pos @punctuation_centered' @geor_lowercase <0 -25 0 0>;
    pos @punctuation_centered' @spaces_all @geor_lowercase <0 -25 0 0>;
    pos @punctuation_centered' @spaces_all @punctuation_centered @geor_lowercase <0 -25 0 0>;
```

Previously, only the first line was executed and the other two were ignored.

```
> hb-shape --text "-ა" build/GoogleSans/variable/GoogleSans\[GRAD,opsz,wght\].ttf
[hyphen=0@0,-25+441|uni10D0=1+506]

> hb-shape --text "- ა" build/GoogleSans/variable/GoogleSans\[GRAD,opsz,wght\].ttf
[hyphen=0+441|space=1+232|uni10D0=2+506]

> hb-shape --text "- -ა" build/GoogleSans/variable/GoogleSans\[GRAD,opsz,wght\].ttf
[hyphen=0+441|space=1+232|hyphen=2@0,-25+441|uni10D0=3+506]
```

With the new fontTools, all are executed:

```
> hb-shape --text "-ა" ../googlesans/build/GoogleSans/variable/GoogleSans\[GRAD,opsz,wght\].ttf
[hyphen=0@0,-25+441|uni10D0=1+506]

> hb-shape --text "- ა" ../googlesans/build/GoogleSans/variable/GoogleSans\[GRAD,opsz,wght\].ttf
[hyphen=0@0,-25+441|space=1+232|uni10D0=2+506]

> hb-shape --text "- -ა" ../googlesans/build/GoogleSans/variable/GoogleSans\[GRAD,opsz,wght\].ttf
[hyphen=0@0,-25+441|space=1+232|hyphen=2@0,-25+441|uni10D0=3+506]
```

So now the code works _as it is written_, at least in bare HarfBuzz (browsers and FontGoggles have the old behavior even with the new fonts, still not sure why). It simply has the side-effect of also shifting the punctuation downwards in `abo- აბო-` because the code has no statements saying to ignore punctuation preceeded by anything non-Georgian.

The question is then: do we care?

razmadzekoko:
The same case would be with Armenian script too, because we decided to have this feature code in Armenian and in Georgian. This was a idea by Gor to make ideal visualisation of shifted symbols because of smaller x-height. 
@chrissimpkins can we ask it to Gor too? If it's so problematic and causes other side effects what should we decide?  If there is no other way and its not working I don't see it so critical to have it but let see the Gor's position too.

chrissimpkins:
> Jany figured it out!!!!

👏 👏 👏 !

chrissimpkins:
> It simply has the side-effect of also shifting the punctuation downwards in abo- აბო- because the code has no statements saying to ignore punctuation preceeded by anything non-Georgian.

Can I interpret this to mean that the hyphen that follows the Latin `abo` also shifts with the feature code as implemented?

> because the code has no statements saying to ignore punctuation preceeded by anything non-Georgian.

If yes to the above question, is it possible to define exclusions to the feature rule?

chrissimpkins:
@gorjious We are discussing the contextual positioning of the hyphen in the Armenian and Georgian sets that will be released next month.  IIUC, the feature code that you and Akaki introduce appropriately shifts the hyphen position when set (with or without an intervening space) next to Armenian/Georgian, but there is a side effect that this also applies to any Latin followed by a hyphen when Armenian or Georgian chars are in the string.  How important is the contextual positioning issue to your Armenian design?

chrissimpkins:
@razmadzekoko @gorjious Does this apply to punctuation other than the hyphen?  I seem to recall that there were other script specific punctuation positioning issues though I think that Nikolaus indicated that the hyphen is the only glyph that fails our shaping tests in the regression text runs.

gorjious:
> @razmadzekoko @gorjious Does this apply to punctuation other than the hyphen? I seem to recall that there were other script specific punctuation positioning issues though I think that Nikolaus indicated that the hyphen is the only glyph that fails our shaping tests in the regression text runs.

I think the original reason for including `pos @punctuation_centered' @spaces_all @geor_lowercase <0 -25 0 0>;` is to account for the combinations that include an `endash`, `emdash`, `bullet` (part of the punctuation_centered class) which can be followed a space. 

That is interesting because from what I know mixing script and features don't work. Khaled Hosny explains it better here: https://forum.glyphsapp.com/t/kerning-multiple-scripts-to-same-punctuation-symbol/15579/2

It was part of an issue before I ran into before: https://github.com/gorjious/google-sans-armenian/issues/5#issuecomment-611025849

If you can't access the repo, here is a screenshot:
Screenshot 2021-06-22 at 19 13 10

I don't think it is a very important issue. It is fine if the select punctuation shifts down when there is Latin in the string with Armenian/Georgian. But it is not a big deal if it doesn't. 

I would be curious to know what is happening in these examples:
`New-York-ը` `New-Yorker-ներ` 
Is the hyphen between Latin and Armenian shifting down? Also, is ն (now-arm) being substituted (now-arm.nonspacing) or is there kerning with the hyphen?

madig:
> Can I interpret this to mean that the hyphen that follows the Latin abo also shifts with the feature code as implemented?

Yes.

> If yes to the above question, is it possible to define exclusions to the feature rule?

Not sure, but also not sure if it's worth it? I agree with Gor... Also, notice how it doesn't seem to happen in browsers. Maybe they segment the text differently?

> though I think that Nikolaus indicated that the hyphen is the only glyph that fails our shaping tests in the regression text runs

No, all punctuation defined in georgian.fea -> `@punctuation_centered` are affected.

> That is interesting because from what I know mixing script and features don't work

Yes, but the code does not look for a different script and punctuation is considered neutral or inheriting whatever is in the current script.

> I would be curious to know what is happening in these examples:

It seems locl is transforming the immediately preceeding hyphen into the Armenian one, but otherwise exhibits the same behavior:

```
> hb-shape --text "Nw-Yk-ը" ../googlesans/build/GoogleSans/variable/GoogleSans\[GRAD,opsz,wght\].ttf
[N=0+698|w=1+777|hyphen=2+401|Y=3+565|k=4+505|uni058A=5@0,-16+441|uni0568=6+550]

> hb-shape --text "Nw-Yk-ներ" ../googlesans/build/GoogleSans/variable/GoogleSans\[GRAD,opsz,wght\].ttf
[N=0+698|w=1+777|hyphen=2+401|Y=3+565|k=4+505|uni058A=5@0,-16+441|uni0576.nonspacing.long=6+550|uni0565=7+541|uni0580=8+550]

> hb-shape --text "Nw-Yk--ներ" ../googlesans/build/GoogleSans/variable/GoogleSans\[GRAD,opsz,wght\].ttf
[N=0+698|w=1+777|hyphen=2+401|Y=3+565|k=4+505|hyphen=5@0,-16+441|uni058A=6@0,-16+441|uni0576.nonspacing.long=7+550|uni0565=8+541|uni0580=9+550]

> hb-shape --text "Nw-Yk- -ներ" ../googlesans/build/GoogleSans/variable/GoogleSans\[GRAD,opsz,wght\].ttf
[N=0+698|w=1+777|hyphen=2+401|Y=3+565|k=4+505|hyphen=5@0,-16+441|space=6+232|uni058A=7@0,-16+441|uni0576.nonspacing.long=8+550|uni0565=9+541|uni0580=10+550]

> hb-shape --text "Nw-Yk-- -ներ" ../googlesans/build/GoogleSans/variable/GoogleSans\[GRAD,opsz,wght\].ttf
[N=0+698|w=1+777|hyphen=2+401|Y=3+565|k=4+505|hyphen=5+441|hyphen=6@0,-16+441|space=7+232|uni058A=8@0,-16+441|uni0576.nonspacing.long=9+550|uni0565=10+541|uni0580=11+550]
```

chrissimpkins:
Let's remove all of the `@punctuation_centered` feature code.  We won't be able to support a feature that has the side effect of repositioning the default punctuation characters in the context within which they are intended to be used.  If I understand this correctly, there is no way to know if this is Georgian/Armenian embedded in a paragraph of Latin-based text or vice versa, and even in an Armenian or Georgian locale, the former would have incorrectly positioned punctuation.

Gor and Akaki, please feel free to open an issue on this if you'd like to explore it further.  Perhaps there is a way to strike a balance in the default form so that it works better across all scripts. 

@madig I assume that this means that we need to revert some of the Armenian fea code as well as the Georgian?

gorjious:
Just to clarify. In a context like this `abo- აბო-` the hyphen after Latin is shifting down? Is this occurring in broswers, FontGoggles, apps?

If the issue is that there should be no case in which punctuation should shift with Latin, regardless if it is embedded in a paragraph of Armenian/Georgian text or not...we can adjust the code so that the feature only occurs when those punctuation marks are surrounded by that script.


```
pos @punctuation_centered' @armn_lowercase <0 -16 0 0>;
pos @punctuation_centered' @spaces_all @armn_lowercase <0 -16 0 0>;
pos @punctuation_centered' @spaces_all @punctuation_centered @armn_lowercase <0 -16 0 0>;
```
By adding the `@armn_lowercase` class before the `@punctuation_centered` this should prevent anything changing for Latin.
```
pos @armn_lowercase @punctuation_centered' @armn_lowercase <0 -16 0 0>;
pos @armn_lowercase @punctuation_centered' @spaces_all @armn_lowercase <0 -16 0 0>;
pos @armn_lowercase @punctuation_centered' @spaces_all @punctuation_centered @armn_lowercase <0 -16 0 0>;
```
We would also need to add the same pattern with the `@armn_uppercase` class 
```
pos @armn_uppercase @punctuation_centered' @armn_lowercase <0 -16 0 0>;
pos @armn_uppercase @punctuation_centered' @spaces_all @armn_lowercase <0 -16 0 0>;
pos @armn_uppercase @punctuation_centered' @spaces_all @punctuation_centered @armn_lowercase <0 -16 0 0>;
```

chrissimpkins:
@gorjious Are you able to [install Harfbuzz with homebrew](https://formulae.brew.sh/formula/harfbuzz) Gor?  I believe that the homebrew installation will give you the `hb-shape` executable so that you can view the output in the examples that Nikolaus was showing with any text run and any font build (in Google Sans or any other project if you want to compare with how other Armenian typefaces address this).

e.g. the Georgian example above,

```
hb-shape --language kat --text "abo- abo- აბო-"
```

This will give you raw shaping output data from the Harfbuzz shaping engine so that you can have a look at text runs of interest.   You'll see glyph subs and shaped metrics.  We use the same approach via a different Python lib based mechanism in our shaping tests.  This is a way to easily assess the same data on the command line.

Docs: https://harfbuzz.github.io/utilities.html#utilities-command-line-hbshape

chrissimpkins:
> @gorjious Are you able to [install Harfbuzz with homebrew](https://formulae.brew.sh/formula/harfbuzz) Gor? I believe that the homebrew installation will give you the `hb-shape` executable so that you can view the output in the examples that Nikolaus was showing with any text run and any font build (in Google Sans or any other project if you want to compare with how other Armenian typefaces address this).
> 
> e.g. the Georgian example above,
> 
> ```
> hb-shape --language kat --text "abo- abo- აბო-"
> ```
> 
> This will give you raw shaping output data from the Harfbuzz shaping engine so that you can have a look at text runs of interest. You'll see glyph subs and shaped metrics. We use the same approach via a different Python lib based mechanism in our shaping tests. This is a way to easily assess the same data on the command line.
> 
> Docs: https://harfbuzz.github.io/utilities.html#utilities-command-line-hbshape

@gorjious Feel free to use a branch in this repository prefixed with `build-` to test any ideas that you have.  e.g., a `build-armenian-tests` branch where you edit fea code will make archives with fonts compiled in our production build workflow that land commit by commit in [this Github Actions thread](https://github.com/googlefonts/googlesans/actions/workflows/upload-artifacts.yml). Choose the commit of interest, download the fonts, and have a look at any experiments with fea code.  

We are running short on time to include significant feature updates that will require review for our v5 release in a few weeks.  We can consider these changes for v6 that will be released at the end of Q3.

chrissimpkins:
@madig Let's remove the contextual punctuation positioning changes in Armenian and Georgian for v5.  We'll wait to hear more from Gor and Akaki about alternative approaches as part of Q3 work targeting GS/GST v6.

gorjious:
Unfortunately, I won't be able to do an install right now. 

That is understandable. Certainly more important to get these updates out there. We can definitely revisit this down the line and maybe figure out a better solution.

chrissimpkins:
> Unfortunately, I won't be able to do an install right now.
> 
> That is understandable. Certainly more important to get these updates out there. We can definitely revisit this down the line and maybe figure out a better solution.

Absolutely no obligation to do anything further Gor (and Akaki). If you ever want to dive into this positioning again, we'd be happy to have a look and consider it in a patch for a future release.

madig:
> Is this occurring in broswers, FontGoggles, apps?

In hb-shape and not those apps, for some reason...

> I assume that this means that we need to revert some of the Armenian fea code as well as the Georgian?

Yes.

> Let's remove the contextual punctuation positioning changes in Armenian and Georgian for v5

Will add commits to this PR tomorrow.
>

madig:
@chrissimpkins to clarify, do you want to jettison the _whole_ `calt` punctuation positioning feature for Armn and Geor? So, `lookup armn_calt_shift_punctuation` for Armn and `feature calt` for Geor?

chrissimpkins:
> @chrissimpkins to clarify, do you want to jettison the _whole_ `calt` punctuation positioning feature for Armn and Geor? So, `lookup armn_calt_shift_punctuation` for Armn and `feature calt` for Geor?

I'll have a look at the Georgian pull request today and let you know

chrissimpkins:
Will post response to the calt issue in the Georgian PR thread.

chrissimpkins:
Auto rebased on main branch following the #310 merge

chrissimpkins:
@madig This gtg?  Please remove the Draft status when you are ready.

madig:
I'll need to update the shaping results to reflect calt removal first.

madig:
Found no binary diffs of fontTools 4.25.0 except the expected backtracking lookup ones vs. main.

chrissimpkins:
> Found no binary diffs of fontTools 4.25.0 except the expected backtracking lookup ones vs. main.

Thanks Nikolaus!

chrissimpkins:
LGTM!

Changelog updated @ d8f642604636e7a72d81b573fdf570b736f6a319

Issue #320, opened 2021-06-18
Refactor Gurmukhi `gur2_dflt_ss01_historical_below_base_consonant_forms` lookup name


chrissimpkins:
We ended up not using `ss01` for this feature in #295.  Let's change the lookup name to `gur2_dflt_historical_below_base_consonant_forms` in the Gurmukhi feature sources in order to avoid confusion down the road.

chrissimpkins:
Closed by #327

PR #327, opened 2021-06-25
Rename historical lookup to remove "ss01"

madig:
Closes https://github.com/googlefonts/googlesans/issues/320.


chrissimpkins (approved):
Ty!

PR #335, opened 2021-07-09
Parallelize build system

madig:
This does some clean up and tightens the Makefiles to make `make` be able to run in parallel.

Static instances could be split up further, but maybe another time. I'm mostly interested in the VFs anyway.

Manually went through the TTX dumps to verify that there are no binary diffs between this and main.


chrissimpkins:
This is great Niko!  Will have a look asap

source/Makefile
```diff
@@ -108,21 +108,18 @@ gs-static: gs-static-upright gs-static-italic
 gs-static-upright: $(FONT_BUILD_DIR)
 	$(FONTMAKE_EXE) -m $(UPRIGHT_SOURCE) --interpolate -o ttf --output-dir $(STATIC_BUILD_DIR) --expand-features-to-instances
 	$(PYTHON3_EXE) ../scripts/gs-subset.py $(STATIC_UPRIGHT_BUILD_PATHS)
-	$(PYTHON3_EXE) ../scripts/set-unicoderange.py $(STATIC_UPRIGHT_BUILD_PATHS)
```
chrissimpkins: This is a convenience feature to eliminate a step to maintain the UFO sources with these data.  Do we really want to / can we remember to update the UFO's with each script expansion?

madig: K! Reverted and going all in by removing Unicode ranges from the UFOs.

chrissimpkins:
Ryzen 3950 / Manjaro 5.4.116-1 / cPy 3.9.5

```
make  595.90s user 3.56s system 99% cpu 10:00.04 total
```

```
make -j 16  601.33s user 4.78s system 350% cpu 2:53.04 total
```

Embarrassingly parallel across the build formats x Roman/Italic styles

:+1: 

I think that we get four cores on the GHA builds?

Unfortunately the stdout becomes garbled, uninterpretable garbage but I think that there is a make flag to collate process stdout strings so that you see the proper process-specific output string sequence on parallelized execution?

chrissimpkins:
Maybe?

```
 -O[TYPE], --output-sync[=TYPE]
                              Synchronize output of parallel jobs by TYPE.

```

chrissimpkins:
> 2:53.04 total

We may be able to get this down into the sub-2 min range with pypy3 #299

madig:
Yes, gmake has a switch to buffer per-subtask output.

Also remember that instance generation could probably be sped up further by first interpolating UFOs and then using all cores to race instance compilation. One day maybe.

Yeah, pypy3 could be something once this PR lands. Would need to go back to making some packages to make it work. I think I got entangled into fontmake cleanup work when I last tried it.

chrissimpkins:
> Also remember that instance generation could probably be sped up further by first interpolating UFOs and then using all cores to race instance compilation. One day maybe.

We could also consider a VF first/only CI workflow with nightly builds/nightly testing of the statics to take them out of the continuous build picture altogether.  If necessary, we could set up a branch filter to run static file tests on demand as needed.  I'm not sure if we absolutely need to build statics on every commit.

madig:
I can add a commit to make statics build on the same schedule as upload-artifacts, i.e. only on `main` and `build*` branches?

chrissimpkins:
Let me give it some thought.  I commonly use the build branches for testing artifacts and don't always need statics.  And waiting until something lands in main to test statics may be too long of an interval.

madig:
Have a look at https://github.com/googlefonts/googlesans/pull/335/commits/73943926247b9d826c959f230a6d27eb85e26bca.

chrissimpkins:
> Yeah, pypy3 could be something once this PR lands. Would need to go back to making some packages to make it work.

Just has this working in our GS CJK builds.  Looking into the approach that he used.

chrissimpkins:
> > Yeah, pypy3 could be something once this PR lands. Would need to go back to making some packages to make it work.
> 
> Just has this working in our GS CJK builds. Looking into the approach that he used.

https://github.com/googlefonts/googlesans/issues/298#issuecomment-888518251

madig:
Rebased on main. Will use as the base for the Thai import.

.github/workflows/static-artifact-ci.yml
```diff
@@ -1,6 +1,14 @@
...
-on: [push]
+on:
+  push:
+    # Sequence of patterns matched against refs/heads
+    branches:
+      - main # Push events on main branch
+      - "build/**" # Push events to branches matching refs/heads/build/[ANYTHING]
+      - "build-*" # Push events to branches matching refs/heads/build-[ANYTHING]
+  # Manual trigger
+  workflow_dispatch:
```
chrissimpkins, line 11: This is nice!  I didn't realize that this was available.

Documentation for future reference: https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/

madig: Me neither. The docs are a bit, uh, unclear on some practicalities.

chrissimpkins: Seems to give us push button access to the static tests from what I can tell?

chrissimpkins:
> Maybe?
> 
> ```
>  -O[TYPE], --output-sync[=TYPE]
>                               Synchronize output of parallel jobs by TYPE.
> ```

Gnu make documentation on parallel output support: https://www.gnu.org/software/make/manual/html_node/Parallel-Output.html#Parallel-Output

.github/workflows/static-artifact-ci.yml
madig: I think so.

chrissimpkins:
It looks like there is very limited output during execution at this stage of the changes that you made so the parallel output support does not seem to be necessary.

madig:
No output is good output.

chrissimpkins:
This looks fantastic Nikolaus.  Thank you very much!

We need to make sure that we either manually, or via a `build-*` branch, run the CI against static instance builds before PR merges going forward.

chrissimpkins:
I never updated the release workflow with your changes.  21 mins and counting waiting on the latest release files... :)

Issue #253, opened 2021-03-04
Add QA check for differences in advance widths and kerning across the GRAD axis masters


chrissimpkins:
This will support the assurance that line length will not change with GRAD axis value changes.

Related https://github.com/googlefonts/googlesans/issues/252

Let's see how this turns out.  It probably belongs upstream if there is a profile that checks UFO's rather than binaries.  If not, we can consider support in upstream fontbakery with a binary testing approach or an entirely separate testing tool.

chrissimpkins:
Per Nikolaus, the approach that we used to import David Berlow's GRAD masters does not allow for differences in kerning across the GRAD masters because we are only taking upstream kerns and replicating across the GRAD masters during the merge process.  It will be useful to add cross-GRAD master kerning checks for regression testing going forward.  Kerning differences should not be an issue for the v4.000 release.

madig:
Kernign can be verified by:

```
nikolaus@DM1248 ~/googlesans (import-berlow-v3)> sha256sum source/GoogleSans/*.ufo/kerning.plist | sort
0920d4dc20ea2f251a9e8bb0ce9baa07e10f4d4d660bd45c153e6ea303fd58b8  source/GoogleSans/GoogleSansItalic-opsz18-wght380-GRAD-50.ufo/kerning.plist
0920d4dc20ea2f251a9e8bb0ce9baa07e10f4d4d660bd45c153e6ea303fd58b8  source/GoogleSans/GoogleSansItalic-opsz18-wght380-GRAD0.ufo/kerning.plist
0920d4dc20ea2f251a9e8bb0ce9baa07e10f4d4d660bd45c153e6ea303fd58b8  source/GoogleSans/GoogleSansItalic-opsz18-wght380-GRAD200.ufo/kerning.plist
0c4669d3cb166e7ceaf598cf28652221336a5cc235a62e8b04e0c98400043b82  source/GoogleSans/GoogleSansItalic-opsz17-wght734-GRAD0.ufo/kerning.plist
219b003f96e298d95af248fc62ac2e45f482fb23b8c0c63f15b84fc57b4b80ec  source/GoogleSans/GoogleSansItalic-opsz18-wght734-GRAD0.ufo/kerning.plist
25faabceea1159ecd576ef5c823727f148fca3577f86b06345fb3d51a46c74d7  source/GoogleSans/GoogleSans-opsz18-wght734-GRAD0.ufo/kerning.plist
[...]

PR #345, opened 2021-07-12
[QA] Add cross-GRAD master kerning checks

chrissimpkins:
Partially addresses #253

### DIFF
--- a/.github/workflows/kerning-ci.yml
+++ b/.github/workflows/kerning-ci.yml
@@ -1,4 +1,4 @@
-name: UFO source kerning checks across GRAD masters
+name: UFO source kerning checks
 
 on: [push]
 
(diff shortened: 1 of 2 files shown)


chrissimpkins:
@madig It looks like we have one set of Hebrew kerns defined in the feature files

https://github.com/googlefonts/googlesans/search?p=1&q=kern

Thoughts about transitioning this to the kerning.plist files so that we can test all cross-GRAD kerning with the kerning.plist SHA hash comparison approach?

madig:
We can't, those are contextual kerns. kerning.plist only carries plain unconditional pair kerning. I think I need to check those again though, because if they are the same across all masters, we can put them into family.fea.

chrissimpkins:
> We can't, those are contextual kerns. kerning.plist only carries plain unconditional pair kerning. I think I need to check those again though, because if they are the same across all masters, we can put them into family.fea.

Ty!  We have these covered in the hb shaping regression tests.  We'll just need to confirm that the kerns are the same across GRAD masters if any new non-kerning.plist defs come in.

PR #350, opened 2021-07-13
[family.fea] Use consistent casing in stylistic set descriptive names

chrissimpkins:
Minor [stylistic set descriptive name](https://adobe-type-tools.github.io/afdko/OpenTypeFeatureFileSpecification.html#8c-descriptive-names-for-stylistic-set-features-ss01---ss20) updates to establish a consistent approach to casing


chrissimpkins:
Confirmed fix in name tables of Roman and Italic builds

PR #367, opened 2021-08-13
Store GDEF information in UFOs

### LINKED ISSUE #112: GDEF: LigPosCaret statements are not being interpolated
Keeping an explicit GDEF table means also keeping an explicit list of LigPosCaret in the features file. Those aren't interpolated however, so instance interpolation will always use the default features and therefore the default LigPosCarets. The results are text carets being off in ligatures in static instances.

### LINKED ISSUE #138: Store GDEF in-UFO instead of in features.fea
Storing GDEF information in features.fea is possible and done today but is also very much annoying. One, it needs separate maintenance and two, it causes https://github.com/googlefonts/googlesans/issues/112.

The UFO spec now allows to generate a complete GDEF at compile time:

1. LigatureCaretByPos can be generated from https://github.com/unified-font-object/ufo-spec/blob/gh-pages/versions/ufo3/glyphs/glif.md#ligature-carets
2. GlyphClassDef can be generated from https://unifiedfontobject.org/versions/ufo3/glyphs/glif/#publicopentypecategory

So, _someone_ needs to sit down and write a GDEF writer for ufo2ft. Since it runs after all (anchor propagation) filters, we might not even need to store any categories. GDEF generation is already done by code in glyphsLib, so that can be used as a start.

### LINKED ISSUE #346: Use `public.openTypeCategories` to organize GDEF table
The per-master GDEF table definition in features.fea is, IIRC, the only real reason we currently have per-master feature files. With the updated pipeline, we can now store that information in a structured form in lib.plist and the compiler will pick it up. Much cleaner, and no need to run the normalizer to update the GDEF definition. Also, maybe correct ligature carets in statics because GDEF is filled in after interpolation?

madig:
Closes https://github.com/googlefonts/googlesans/issues/346
Closes https://github.com/googlefonts/googlesans/issues/138
Closes https://github.com/googlefonts/googlesans/issues/112

This makes maintenance and mergers easier.

TTX diffs show no diff for VFs and new ligature coordinates in static instances.


chrissimpkins:
D/W Nikolaus today. This is ready to go and can be merged before we merge the Thai script.

Issue #351, opened 2021-07-14
GPOS file size optimizations 2


chrissimpkins:
From @simoncozens:

> The idea is to use a custom mark feature writer than quantises anchors to the nearest 10 units. That allows more anchor tables to be shared.  I'm not quite sure what the right place for it is. It is a general purpose tool, but requires setting custom parameters. I'll make a ufo2ft PR in the first instance.

~~https://github.com/googlefonts/ufo2ft/pull/513~~
https://github.com/googlefonts/ufo2ft/pull/516

chrissimpkins:
@simoncozens How do you define your new approach (https://github.com/googlefonts/ufo2ft/pull/516) in UFO sources so that these changes are used in a fontmake pipeline?

simoncozens:
Put this in the top-level `<lib>` element of your designspace file:

```xml
      <key>com.github.googlei18n.ufo2ft.featureWriters</key>
      <array>
        <dict>
          <key>class</key>
          <string>MarkFeatureWriter</string>
          <key>options</key>
          <dict>
            <key>mode</key>
            <string>skip</string>
            <key>quantization</key>
            <integer>8</integer>
          </dict>
        </dict>
      </array>
```

chrissimpkins:
#386

chrissimpkins:
Will push these edits back to the v9.000 release per Fonts eng request

khaledhosny:
FWIW, if file size really a concern, you can save ~40KB from each VF font by dropping glyph names (e.g. using version 3 post table).

madig:
https://github.com/googlefonts/googlesans/issues/115

khaledhosny:
> #115

Thanks, I missed that one.

belluzj:
I'm reposting here a comment from Chris in the GSF engineering chat, replying to file size saving reports here: https://github.com/googlefonts/googlesans/pull/386#issuecomment-1768137624

> Thanks for the additional data on the potential max file size savings.  I think that we abandon this mark quant effort in favor of other higher priority areas.  I don't see that level of file size savings impacting decisions to land the family, and it is bound to make it more challenging to land because it pixel-level shifts marks.  Worth having a look at this in GS Flex, though.  We're much more open to change in GS Flex v2 and it would be interesting to understand if the impact happens to be greater for some reason there.

So this can be closed, but the same issue should be opened on GS Flex, right?

PR #386, opened 2021-09-15
Use mark quantization filter

madig:
Test PR for https://github.com/googlefonts/googlesans/issues/351.

### DIFF
--- a/source/GoogleSans/GoogleSans-opsz17-wght380-GRAD-50.ufo/lib.plist
+++ b/source/GoogleSans/GoogleSans-opsz17-wght380-GRAD-50.ufo/lib.plist
@@ -14,7 +14,7 @@
         <key>options</key>
         <dict>
           <key>quantization</key>
-          <integer>8</integer>
+          <integer>32</integer>
         </dict>
       </dict>
       <dict>
--- a/source/GoogleSans/GoogleSans-opsz17-wght380-GRAD-50.ufo/lib.plist
+++ b/source/GoogleSans/GoogleSans-opsz17-wght380-GRAD-50.ufo/lib.plist
@@ -2,6 +2,26 @@
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
   <dict>
+    <key>com.github.googlei18n.ufo2ft.featureWriters</key>
+    <array>
+      <dict>
+        <key>class</key>
+        <string>KernFeatureWriter</string>
+      </dict>
+      <dict>
+        <key>class</key>
+        <string>MarkFeatureWriter</string>
+        <key>options</key>
+        <dict>
+          <key>quantization</key>
+          <integer>8</integer>
+        </dict>
+      </dict>
+      <dict>
+        <key>class</key>
+        <string>GdefFeatureWriter</string>
+      </dict>
+    </array>
     <key>com.github.googlei18n.ufo2ft.filters</key>
     <array>
       <dict>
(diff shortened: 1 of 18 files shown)


madig:
Ok, it trips a whole lot of regression tests, I guess that's expected and I'd need to dig into the diffs. But it also kills the `dist` feature?!

Edit: I messed up feature writer options.

madig:
Ok. Now it shaves off 13 KB off both variable fonts and 3-4 KB off the statics.

chrissimpkins:
> Ok. Now it shaves off 13 KB off both variable fonts and 3-4 KB off the statics.

cc @simoncozens

chrissimpkins:
Hoo boy.  Lots of HB shaping regression test breakage.  I'm not sure where to begin visually testing this.

simoncozens:
Makes sense, since the whole point is to adjust the anchor positions!

chrissimpkins:
@simoncozens have you done any visual testing with this feature?

simoncozens:
I use a similar technique in Fez, and it works fine. If you look at the regression tests, the positioning shouldn't be more than (in your case) 8 units out from the test values in either direction.

source/GoogleSans/GoogleSans-opsz17-wght380-GRAD-50.ufo/lib.plist
```diff
@@ -20,6 +20,8 @@
         <dict>
           <key>mode</key>
           <string>skip</string>
+          <key>quantization</key>
+          <integer>8</integer>
```
chrissimpkins: Can you remind me of the rationale behind the use of the value 8 here?

simoncozens: Behdad suggested that it gives bit representations which compress nicely with WOFF etc.

madig:
So, this saves 33 KB on the upright VF and 53 KB on the italic VF. After woff2-compressing them, the diff is 25 KB and 15 KB respectively.

belluzj:
The build is failing; if we want to go ahead with this approach and merge we'll need first to re-generate the shaping expectations (because every mark has moved a bit)

chrissimpkins:
Can we think of any way to make the file size savings more significant relative to our overall file size?  IIUC this is a 25kb improvement which represents about 0.5% savings relative to the overall file size.  And with the extensive set of changes involved at the pixel diff level this is going to make it more difficult to roll this family out into production.  IMO ok to close unless we can find a way to increase the file size savings.

Hoolean:
Jany suggested that we try building with a high but impractical quantization tolerance (e.g. 64+) just to see what the upper bound on improvements here are, I will push a commit that does this now

Hoolean:
@MariannaPaszkowska @chrissimpkins

The results are in! 🥁

The values for `quantisation=32` provide an upper bound for file size savings from the filter; in practice, we would never want to apply a filter that coarse on all glyphs, and so real-world savings would be less.

In kilobytes:

| File | Original | `quantization=8` | `quantization=32` |
|-|-:|-:|-:|
| GoogleSans[GRAD,opsz,wght].ttf | 4740 | (-33, 99.3%) 4707 | (-59, 98.8%) 4681 |
| GoogleSans[GRAD,opsz,wght].woff2 | 1460 | (-14, 99.0%) 1446 | (-28, 98.1%) 1432 |
| GoogleSans-Italic[GRAD,opsz,wght].ttf | 5034 | (-53, 98.9%) 4981 | (-82, 98.4%) 4952 |
| GoogleSans-Italic[GRAD,opsz,wght].woff2 | 1680 | (-25, 98.5%) 1655 | (-38, 97.7%) 1642 |

PR #391, opened 2021-09-22
Delete Greek ccmp entries

### LINKED ISSUE #195: Remove Greek part of ccmp
https://github.com/googlefonts/googlesans/issues/161#issuecomment-777796745

madig:
Implements https://github.com/googlefonts/googlesans/issues/161#issuecomment-777796745.

This removes the `.case` variants of various Greek marks (without removing them from the source). I am not sure this is as it should be? @moyogo @irenevl


chrissimpkins:
Thoughts about this @irenevl and @moyogo?  OK to merge as part of the v6.000 release scheduled next week?

irenevlachou:
> Thoughts about this @irenevl and @moyogo? OK to merge as part of the v6.000 release scheduled next week?

I think that's ok, we just need them for composing the accented uc, not as standalone glyphs

madig:
Ok, then we can leave them in. We flatten components, so the source glyphs can use the `.case` variants and they get removed by the subsetter because the production glyphs use the transformed outlines directly.

chrissimpkins:
Thanks Irene! Thanks Nikolaus!

PR #409, opened 2021-10-20
Add Fontbakery configuration

simoncozens:
Please review commit-by-commit. The changes in the first three commits should be uncontroversial. The last change is a bit more interesting.

Fontbakery has recently grown both a configuration file mechanism and a way of providing constant values to checks (similar to your `ATTRIBUTES` dictionary), but which can be [overridden by user configuration](https://github.com/googlefonts/fontbakery/pull/3427/commits/86080ff7753c388371cebcd2376c53964787addf).

The canonical example is checking for file size. The GF profile specifies a default value to warn/fail at:

```python
    "com.google.fonts/check/file_size": {
         "WARN_SIZE": 1 * 1024 * 1024,
         "FAIL_SIZE": 9 * 1024 * 1024
     }
```

But if I am working on a CJK family, I might want to put this in my `fontbakery.yml`:

```yaml
com.google.fonts/check/file_size:
  WARN_SIZE: 10485760
  FAIL_SIZE: 31457280
```

The last commit replaces the `ATTRIBUTES` directory with these check-specific constants. The reason for doing this is that "UI builds" of GS and certain script-specific font builds may need to deviate from some of the values - for example, underline values in the Myanmar language-specific builds. With this mechanism in place, developers of specific font builds can add a configuration file to override those values.

### DIFF
--- a/qa/check-googlesans.py
+++ b/qa/check-googlesans.py
@@ -67,24 +67,46 @@ excluded_check_ids = (
     "com.adobe.fonts/check/family/consistent_family_name",
 )
 
-ATTRIBUTES = {
-    "os2_fsselection_bit7": 1,
-    # "ymax": 1115,  # defined at max across min + max opsz design space (from min opsz)
-    # "ymin": -292,  # defined at min across min + max opsz design space (from min opsz)
-    "hhea_ascent": 966,  # set to match typo metrics values
-    "hhea_descent": -286,
-    "hhea_linegap": 0,
-    "os2_strikeout_position": 306,
-    "os2_strikeout_size": 84,
-    "os2_typoascender": 966,  # set to match hhea metrics values
-    "os2_typodescender": -286,
-    "os2_typolinegap": 0,
-    "expected_fvar_axes": ["opsz", "wght", "GRAD"],
-    "opsz_axis_default": 18.0,
-    "post_underline_position": -160,
-    "post_underline_thickness": 84,
-    "wght_axis_default": 400.0,
-    "grad_axis_default": 0.0,
+profile.configuration_defaults = {
+    "com.google.fonts/check/googlesans/opentype/os2/fsselectionbit7": {
+        "os2_fsselection_bit7": 1
+    },
+    # "com.google.fonts/check/googlesans/opentype/os2/winascent": {
+    #     "os2_win_ascent": 1115  # defined at max across min + max opsz design space (from min opsz)
+    # },
+    # "com.google.fonts/check/googlesans/opentype/os2/windescent": {
+    #     "os2_win_descent": 292  # defined at min across min + max opsz design space (from min opsz)
+    # },
+    "com.google.fonts/check/googlesans/opentype/hhea/ascent": {
+        "hhea_ascent": 966  # set to match typo metrics values
+    },
+    "com.google.fonts/check/googlesans/opentype/hhea/descent": {
+        "hhea_descent": -286  # set to match typo metrics values
+    },
+    "com.google.fonts/check/googlesans/opentype/hhea/linegap": {"hhea_linegap": 0},
+    "com.google.fonts/check/googlesans/opentype/os2/strikeout": {
+        "os2_strikeout_position": 306,
+        "os2_strikeout_size": 84,
+    },
+    "com.google.fonts/check/googlesans/opentype/os2/typodescender": {
+        "os2_typodescender": -286,
+    },
+    "com.google.fonts/check/googlesans/opentype/os2/typoascender": {
+        "os2_typoascender": 966,  # set to match hhea metrics values
+    },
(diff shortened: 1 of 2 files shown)


simoncozens:
Obviously flake8 freaks out about the variable injection.

qa/check-googlesans.py
```diff
@@ -58,26 +58,46 @@
...
-    "os2_typoascender": 966,  # set to match hhea metrics values
-    "os2_typodescender": -286,
-    "os2_typolinegap": 0,
-    "expected_fvar_axes": ["opsz", "wght", "GRAD"],
-    "opsz_axis_default": 18.0,
-    "post_underline_position": -160,
-    "post_underline_thickness": 84,
-    "wght_axis_default": 400.0,
-    "grad_axis_default": 0.0,
+profile.configuration_defaults = {
```
chrissimpkins, line 70: Can these fields be specified in an external yaml/toml config file that derivative branches can then modify?  Having an easy to edit upstream config file format available at the time that they branch might be an easier approach for some development teams.

qa/check-googlesans.py
```diff
@@ -144,10 +164,11 @@
     Confirms that fonts have OS/2.fsSelection bit 7 (USE_TYPO_METRICS) set \
     for typo vertical metrics (instead of win vertical metrics)
     """,
+    configs=["os2_fsselection_bit7"],
```
chrissimpkins, line 179: All config fields are pooled into a single collection by dict key string?  Is the unique check ID relevant in this configuration approach?

chrissimpkins (commented):
Thanks for the fixes Simon!  We can include the fix commits immediately if you want to split them out into a separate PR.

Will you point me to documentation for the new configuration approach?  I'd like to understand the external config file location/name/format and how to expand the available custom checks with the configuration specified in the Python check profile.

chrissimpkins:
> Obviously flake8 freaks out about the variable injection.

Np, we'll sort it out.  We can either inline noqa ignore them or split out the flake8 check on the FB profile with a global ignore.

chrissimpkins:
I cherry picked your fb custom profile fix commits in #410 and will rename this PR "Add fontbakery configuration".

chrissimpkins:
Merged your fixes. This PR is ready to be rebased

qa/check-googlesans.py
chrissimpkins: > Can these fields be specified in an external yaml/toml config file that derivative branches can then modify?

@simoncozens is this possible?

chrissimpkins:
I'll postpone this to v8.  There are still open questions about the approach here will work for downstream teams.

chrissimpkins:
@felipesanches Felipe, is it possible to define an upstream fontbakery configuration file in this project (e.g., yaml, toml, etc) that a downstream team working on a derivative can revise according to their needs?  I'm trying to understand how the FB configs work to determine how to approach Simon's changes here.

felipesanches:
@chrissimpkins, I want to review this, but it will probably take me at least a few more days to be able to do so. I'll cut a new fontbakery release before coming back here to review this PR.

chrissimpkins:
> @chrissimpkins, I want to review this, but it will probably take me at least a few more days to be able to do so. I'll cut a new fontbakery release before coming back here to review this PR.

No rush at all Felipe.  Ty!

qa/check-googlesans.py
simoncozens: Sorry, didn’t reply to this: This is exactly what this PR does - establishes default configuration values which can be overridden in a YAML file. Have a look at the explanation at the top again. :-)

chrissimpkins: My question is if it is possible to define that YAML config file in this upstream repository so that downstream teams only need to modify values in an existing file when they branch, rather than trying to figure out how to spec the configuration file from scratch.

chrissimpkins:
Note to self: docs on Fontbakery config files https://font-bakery.readthedocs.io/en/stable/user/USAGE.html#configuration-file

qa/check-googlesans.py
simoncozens: Sure, yes. Just set up a fontbakery.yml file in upstream. The format should be the YAML equivalent of the `configuration_defaults`. i.e.:

```
com.google.fonts/check/googlesans/opentype/os2/winascent:
         os2_win_ascent: 1115
...
```

and supply `--config fontbakery.yml` on your fontbakery command line.

RickyDaMa:
Rebased this PR, now it's just the last of the four original commits (as the others have already been applied to main). I also kept the implementation of `com_google_fonts_check_googlesans_opentype_os2_winascent` from `main` as we're not using constants anymore (and using them results in Fontbakery fails)

Update: I've also nudged the code a bit and added a `flake8` exception for `check-googlesans.py` so it passes happily

MariannaPaszkowska:
Many thanks, everyone! 🚀

PR #410, opened 2021-10-22
Fontbakery fixes

chrissimpkins:
Pulls in Simon's fontbakery custom profile fixes in #409 to separate them from the new fontbakery configuration approach


chrissimpkins:
Static font build CI run in https://github.com/googlefonts/googlesans/actions/runs/1374034976

PR #413, opened 2021-11-25
Lao

cadsondemakoffice:


chrissimpkins:
Import info from Kano:
[GS package.zip](https://github.com/googlefonts/googlesans/files/7613786/GS.package.zip)

PR #415, opened 2021-11-25
Khmer

cadsondemakoffice:


chrissimpkins:
Import info from Kano:
[GS package.zip](https://github.com/googlefonts/googlesans/files/7613784/GS.package.zip)

PR #425, opened 2021-12-16
Fix graded width of various Greek polytonic glyphs

madig:
I ran a script over the sources that simply set the GRAD0 width on the -50 and 200 master Polytonic Greek glyphs.

---

### TODO

- [x] Irene review of the auto-updated designs
- [x] remove copied cPython source and squash before we merge this
- [x] document that the new Python script requires Py3.9+

PR #417, opened 2021-11-29
Import Khmer

madig:
Closes #415.
Closes #446

---

@chrissimpkins edits below:

- test string for tall / deep glyph testing: `ប៊ិនទៅកម្មវិធីលក់នំធ្វើពីផ្លែប៊័រ   ក្នុងទឡ្ហីករណ៍ដែល​នាង​លក្ស្មី​បង្ហាញហ្វ្លូ`

---

### TODO

- [x] add Apache license to all new Python scripts in `scripts/khmer-import` dir
- [x] open issue on below base vowel sign collision in bold https://github.com/googlefonts/googlesans/issues/453
- [x] open issue on base form + vowel sign ligature separation issue in bold opsz 17 (https://github.com/googlefonts/googlesans/issues/454)

PR #416, opened 2021-11-29
Import Lao

madig:
Closes https://github.com/googlefonts/googlesans/pull/413.

---

@chrissimpkins edits below:

- Test string for tall/deep glyph testing: `ກໍ່ກໍ່ກໍ໊ກໍ໋ກັ່ກັ້ກັ໊ກັ໋ ງູຫຼູ`

PR #429, opened 2021-12-23
Update requirements (based on v6.000)

belluzj:


chrissimpkins:
Builds failed with fontmake update

chrissimpkins:
It looks like we may have some bracket layer / rvrn glyph definitions still around?

> ERROR:fontmake.compatibility:Fonts had differing number of contours in glyph y.BRACKET.18

madig:
Seems like a fontmake issue: https://github.com/googlefonts/fontmake/pull/832#issuecomment-1004936839

chrissimpkins:
> Seems like a fontmake issue: [googlefonts/fontmake#832 (comment)](https://github.com/googlefonts/fontmake/pull/832#issuecomment-1004936839)

Looks like maybe fixed in https://github.com/googlefonts/fontmake/pull/840?

If so, I guess we'll need a new fontmake release to push Khmer with the sources as is. Or we can remove the non-exported glyphs that are causing problems?

belluzj:
@chrissimpkins when is your target release date for v7.000? @anthrotype and @madig are both back next week for the new fontmake release.

Otherwise I could go and remove the BRACKET glyphs from the sources.

chrissimpkins:
> @chrissimpkins when is your target release date for v7.000? @anthrotype and @madig are both back next week for the new fontmake release.
> 
> Otherwise I could go and remove the BRACKET glyphs from the sources.

Jany and I discussed this issue during a meeting today.  We will push the Khmer import that requires the fontmake update back to the GS v8 release so that we have sufficient time to work through the ability to compile Khmer and complete the review process on production builds that use the updated pipeline.

madig:
Looking at the diffs pre/post update, I found:

1. Various lookups turned into extension lookups (probabaly the change in fontTools 4.28.4, https://github.com/fonttools/fonttools/pull/2465)
2. Various GSUB glyph coverage records dropped duplicate entries (change in fontTools 4.28.2, https://github.com/fonttools/fonttools/issues/2433)

So the update looks good to me.

madig:
@chrissimpkins how should we proceed here? I think the update can go in, and then we can rebase #425, #416 and #417 on top?

chrissimpkins:
IIUC we don't need this for #435 or #416.  I think that I will release those changes without the fontmake update, then we can base the Khmer work in #417 (and everything else planned for GS v8.000) on the updated compiler. Sg?

madig:
Ok :)

chrissimpkins:
> Ok :)

Sorry! I need a pause in changes so that I can prep GS v7 this week. :) 

Keep forging ahead on the Khmer.  We'll try to wrap up that import, build, and review over the next couple of weeks before we dive into the next script.

madig:
This is ready to go.

Issue #419, opened 2021-12-06
Update all Postscript names


madig:
Passing the list of glyph names through Glyphs.app, I find various different production names. We should update them wholesale with whatever Glyphs.app says probabaly. glyphsLib inference is unfortunately unreliable.

anthrotype:
You should file an issue against glyphsLib detailing the differences. It may be a different GlyphData.xml or some missing logic or a bug

madig:
Also: there are diffs between upright and italic production names. Groan.

chrissimpkins:
> Also: there are diffs between upright and italic production names. Groan.

Is the Roman v. Italic difference a glyphsLib issue?  or this is coming from definitions that design teams used in sources?

madig:
I'm not sure, It might also be a slip-up. The glyphsets of upright and itralic are not equivalent so there's room for diffs.

madig:
Reported at https://github.com/googlefonts/glyphsLib/issues/770 and Khaled already submitted a PR :)

The openTypeCategory detection needs a looking at next, but that's a separate issue.

chrissimpkins:
Ty!

PR #412, opened 2021-11-02
Import cap height default / international figures in opsz 18 masters

madig:
Taken from https://github.com/googlefonts/googlesans/tree/berlow-figures.

Imported 

- numbersign 
- dollar 
- percent 
- zero 
- one 
- two 
- three 
- four 
- five 
- six 
- seven 
- eight 
- nine 
- cent 
- sterling 
- yen 
- One-roman 
- Two-roman 
- Three-roman 
- Four-roman 
- Five-roman 
- Six-roman 
- Seven-roman 
- Eight-roman 
- Nine-roman 
- Ten-roman 
- Eleven-roman 
- Twelve-roman 
- Fifty-roman 
- Hundred-roman 
- Fivehundred-roman 
- Thousand-roman 
- Onethousandcd-roman 
- Fivethousand-roman 
- Tenthousand-roman 
- numbersign.tf 
- baht 
- dong 
- euro 
- florin 
- franc 
- hryvnia 
- lira 
- liraTurkish 
- peso 
- ruble 
- rupeeIndian 
- tenge 
- tugrik 
- won 
- baht.cap 
- cent.cap 
- cent.tf 
- currency.tf 
- dollar.tf 
- dong.tf 
- euro.tf 
- florin.tf 
- franc.tf 
- hryvnia.tf 
- lira.tf 
- liraTurkish.tf 
- peso.tf 
- ruble.tf 
- rupeeIndian.tf 
- sterling.tf 
- tenge.tf 
- tugrik.tf 
- won.tf 
- yen.tf 
- degree 
- pertenthousand 
- perthousand 
- centigrade 
- fahrenheit 
- sheqel 
- sheqel.tf 
- zeropercent

---

@chrissimpkins edits below:

Text runs for testing:

```
0123456789 
#%‱‱°℃℉
ⅠⅡⅢⅣⅤⅥⅦⅧⅨⅩⅪⅫⅬⅭⅮⅯↀↁↂ
$¢£¥฿₫€ƒ₣₴₤₺₱₽₹₸₮₩₪
```
---

Issue #375, opened 2021-08-31
Add git commit short SHA hash string in name table ID 5 record


chrissimpkins:
Request from Fonts Eng.  Add the git commit SHA hash string to our nameID 5 record as part of our GHA post-compile edit workflow.  This will be used as a link back to the source state across the various internal distribution approaches.

PR #441, opened 2022-01-19
Clean up sources

### LINKED ISSUE #368: Consistent glyph naming guidance & updates
Thanks! So glyphs like `nikhahit_maiEk-thai` (listed under "additional glyphs" in https://github.com/googlefonts/googlesans/pull/361#issuecomment-889869092) are left named like that. Ideally they are named like `uni0E4D_uni0E48`.

@chrissimpkins this is currently handled inconsistently across scripts... not sure if we should retroactively change this, as e.g. Adobe apps may rely on old names when users embedded them into documents by accident...

_Originally posted by @madig in https://github.com/googlefonts/googlesans/issues/361#issuecomment-898557960_

### LINKED ISSUE #418: Remove `<note>` elements
I see a lot of junk hanging around.

### LINKED ISSUE #420: Switch MarkFeatureWriter to append mode
It's in skip right now and since we have no explicit mark or mkmk feature, it never skips, but it would be better to have in append mode by default.

### LINKED ISSUE #430: Source check: hunt down floating point coordinates
We don't want floating point coordinates in the sources because we target TTFs, where we have to round anyway. Sources should also store coordinates like `123.0` as `123`.

madig:
1. Remove all `<note>` elements, for they carried no interesting information
2. Add script to normalize all integer floats coordinates (e.g. 123.0) to actual integers (e.g. 123),
3. Update production names with what Glyphs.app thinks they should be. Various names slipped through the cracks in earlier script merges. Also update expectations with new names.
4. Move the `com.github.googlei18n.ufo2ft.featureWriters` keys from the UFO libs to the Designspace libs, because fontmake copies them where they need to be and we reduce data duplication. Also switch the mark writer to append mode so we can generate mark features and mix them with mark features in the feature file.

Closes #418
Closes #430
Closes #368
Closes #420

---

### TODO

- [x] address question about automated note removal normalization of sources in https://github.com/googlefonts/googlesans/pull/441#issuecomment-1021479223
- [x] float normalization issue: discuss https://github.com/googlefonts/googlesans/pull/441#discussion_r792026744
- [x] address question about how arrived at postscript name updates in https://github.com/googlefonts/googlesans/pull/441#issuecomment-1021542236
- [x] address question about how we address previous names in downstream files in https://github.com/googlefonts/googlesans/pull/441#issuecomment-1021545983
- [x] discuss build commit SHA in json shaping def files in https://github.com/googlefonts/googlesans/pull/441#discussion_r792072029
- [x] discuss why there are new PS names defined (https://github.com/googlefonts/googlesans/pull/441#issuecomment-1025882135)
- [x] post glyph PS naming "bug" to glyphsLib repo (Cosimo request) (https://github.com/googlefonts/googlesans/pull/441#issuecomment-1025805469) - tracking in #419 

### Review

@chrissimpkins LGTM:

- [x] remove note elements
- [x] normalize string formatted integer float values to string formatted integer values
- [x] update all PS names
- [x] consistent glyph naming guidance and updates
- [x] switch mark feature writer to append mode


chrissimpkins:
The glyph name changes here will move us away from the glyph names used in downstream sources by development teams, correct?

madig:
What do you mean? The design names stay as they are, what changes is that the production names in the final TTFs are what Glyphs.app would name them.

chrissimpkins:
> What do you mean? The design names stay as they are, what changes is that the production names in the final TTFs are what Glyphs.app would name them.

OK, I will have a closer look.  6500 files diff so not sure where to begin :)

madig:
Yes it's brutal. Maybe per commit? And then diff and look for patterns? Maybe you can filter out lines not starting with plus or minus.

We need a format and data organization where this kind of stuff does not happen.

belluzj:
@chrissimpkins What you could do as well is review Nikolaus's Python scripts that he used to generate the changes; if you agree with the script then you agree with the changes (and the script should be small enough to review). Please ask if you'd like @madig to re-organize the commits to have the scripts and the data changes separate, or better indicate which commits need review and which commits are merely the application of a script

chrissimpkins:
Will you please document what you changed in the lead post on this PR?  It will help to have a summary of what changes you included and what you are attempting to do to address the issues that you are closing here.

It looks like you've based the Khmer and Ethiopic imports on these edits.  I'll begin to have a look at this branch today. In the future, let's try to avoid massive multi-issue PR's.

chrissimpkins:
Re: remove note elements #418, commit https://github.com/googlefonts/googlesans/pull/441/commits/378a6a811890cbcd424dd387a557cd041355027f

Do we want to include this as part of our source normalization process and run this before notes ever land in the production sources?  My preference would be to *not* use UFO notes for annotations and this would avoid future rounds of note cleanup.  Designers can use them for local work but that will not land in production source and will not round trip back to them from upstream prod sources.

source/GoogleSans/GoogleSans-opsz17-wght380-GRAD-50.ufo/glyphs/tt_tta-deva.glif
```diff
@@ -1,9 +1,6 @@
 <?xml version='1.0' encoding='UTF-8'?>
 <glyph name="tt_tta-deva" format="2">
   <advance width="521"/>
-  <note>
-Missing anchors
```
chrissimpkins, line 5: @kalapi We are removing note elements from the production UFO sources and came across a few Devanagari glyphs that include notes about missing anchors.  I assume that these are dated, irrelevant notes? Pinging you here so that you can have a look through them to confirm that there is nothing actionable.

chrissimpkins: You can review commit 378a6a811890cbcd424dd387a557cd041355027f

madig:
> Do we want to include this as part of our source normalization process and run this before notes ever land in the production sources? 

I thougth about this but am not sure. You could arguably have interesting things in notes but I have never seen them used for anything except review stuff.

chrissimpkins:
> > Do we want to include this as part of our source normalization process and run this before notes ever land in the production sources?
> 
> I thougth about this but am not sure. You could arguably have interesting things in notes but I have never seen them used for anything except review stuff.

Let's see if any designers weigh in with notes workflows where they are valuable.

chrissimpkins:
> > > Do we want to include this as part of our source normalization process and run this before notes ever land in the production sources?
> > 
> > 
> > I thougth about this but am not sure. You could arguably have interesting things in notes but I have never seen them used for anything except review stuff.
> 
> Let's see if any designers weigh in with notes workflows where they are valuable.

It looks like the vast majority, if not all, of the note elements were in the Devanagari work. We either purged the notes during other imports or other designers are not using source file notes.

scripts/internal/normalize.py
```diff
@@ -240,6 +268,14 @@ def scrub_ufo(
     ufo.lib["public.openTypeCategories"] = ot_categories
 
 
+def clean_number(v: Union[int, float]) -> float:
+    if isinstance(v, int):
+        return v
+    if v.is_integer():
```
chrissimpkins, line 274: Got it.  So we only eliminate the decimal representation in the string formatted float *if* it meets the Python `float.is_integer()` criterion (i.e. `1.0` will be cleaned to `1`, `1.5` will remain `1.5`).

chrissimpkins: re: this comment in the original issue report

> We don't want floating point coordinates in the sources because we target TTFs, where we have to round anyway. 

We are not doing this in this PR, correct? If there is anything other than a zero following the decimal point, then it stays with these changes?  I don't see any other int casts, rounds, ceils, floors, etc.

chrissimpkins: IIUC, the signature may need to be the following? (different return type):

```python
def clean_number(v: Union[int, float]) -> Union[int, float]
```

madig: > We are not doing this in this PR, correct? 

Yes. This change will also help make actual floats stand out more :)

> IIUC, the signature may need to be the following? (different return type):

Nope, `float` encompasses `int`, see https://mypy.readthedocs.io/en/latest/duck_type_compatibility.html

chrissimpkins: > Nope, float encompasses int, see https://mypy.readthedocs.io/en/latest/duck_type_compatibility.html

Confusing but ok... :)

This issue makes me wonder why issues like https://github.com/googlefonts/googlesans/pull/412#issuecomment-1011476560 are happening in serialization pipelines.  Odd.

chrissimpkins: This would be a nice write option in ufolib2 (and other UFO I/O libs...) :)

chrissimpkins:
Re: normalization of string formatted float integer values (#430, commits 7fabe3650f3879942243460800bc277f25d0fbd1 & 7d05f531515c5dbdcd6dafc0e9189d3c0ec85710)

My questions in https://github.com/googlefonts/googlesans/pull/441/files/378a6a811890cbcd424dd387a557cd041355027f..7fabe3650f3879942243460800bc277f25d0fbd1#r792026744 were addressed.  I spot checked the glif file edits and the script appears to be doing what we expect it to do across positive and negative coordinate values. This part LGTM.

chrissimpkins:
@madig 

Re: update PS names #419

Can you describe how you arrived at the edits in https://github.com/googlefonts/googlesans/pull/441/commits/c7e489093640193cadd8f9fe6068f9b5d6408df9?  I see the following comment in your commit:

> Using Glyphs 2.6.6.

chrissimpkins:
Re: update PS names #419

Also, downstreams will still have the previous prod glyph names defined I think?  How does this impact our work with downstream teams?  We pull the data once (only), update according to the process used here, and then never overwrite upstream lib.plist data from those downstream files again?  Maybe we should distribute the approach that you used here to every downstream team so that they can do this *before* they push sources to us and keep the downstream in line with the upstream definitions.

_Edit: this was what I was driving at in https://github.com/googlefonts/googlesans/pull/441#issuecomment-1017794570_

qa/shaping/armenian_case.json
```diff
@@ -1,6 +1,6 @@
 {
   "meta": {
-    "build_commit": "63559902ed6174bf979b08fde7b6ddcec9bd9eac"
+    "build_commit": ""
```
chrissimpkins: I suggest that we update the script to record the commit during execution rather than abandon the build commit record in the file.  I don't believe that the commit of last json file change accurately represents the state of the uncommitted artifacts.  Your changes *could* be against files that were compiled in an entirely different branch or those that have been sitting around for weeks as new commits land.  Another idea is that we can just record the name table version string in this field.  We record the artifact build source commit SHA in the name table as of #375.  Then we don't need to manually address it following the json file updates.

madig:
> Can you describe how you arrived at the edits in [c7e4890](https://github.com/googlefonts/googlesans/commit/c7e489093640193cadd8f9fe6068f9b5d6408df9)? I see the following comment in your commit:
> 
> > Using Glyphs 2.6.6.

I copy-pasted the names of all glyphs into a text file (for uprights and italics), one glyph per line, ssh-copied it over to a Mac and ran the following script inside Glpyhs.app:

```python
import pprint
import logging

with open("/tmp/glyphlist_italic.txt") as f:
    names = [n.strip() for n in f.readlines() if n]

production_names = {}

for name in names:
    i = Glyphs.glyphInfoForName(name)
    if i is None:
    	logging.warning("No info for %s", name)
        continue
    if i.productionName is not None and name != i.productionName:
        production_names[name] = i.productionName

pprint.pprint(production_names)
```

I then copy-paste the results out and insert them into a Python session and do 

```
In [1]: for p in Path("source/GoogleSans/").glob("*.ufo"):
   ...:     if "Italic" in p.name:
   ...:         continue
   ...:     u = Font.open(p)
   ...:     u.lib["public.postscriptNames"].update(up)
   ...:     u.save()

In [2]: for p in Path("source/GoogleSans/").glob("*Italic*.ufo"):
   ...:     u = Font.open(p)
   ...:     u.lib["public.postscriptNames"].update(it)
   ...:     u.save()
```

Isn't it wonderful.

madig:
> Also, downstreams will still have the previous prod glyph names defined I think? How does this impact our work with downstream teams? We pull the data once (only), update according to the process used here, and then never overwrite upstream lib.plist data from those downstream files again? Maybe we should distribute the approach that you used here to every downstream team so that they can do this before they push sources to us and keep the downstream in line with the upstream definitions.

Downstreams will have old Postscript names but it doesn't matter, because they are only used in the final TTFs. We only import new glyphs from downstreams, so old ones should not be overwritten. Downstreams can probably embed the production names directly by running a script in Glyphs.app, though I'd need to find one. What I'm doing is simply to paper over glyphsLib <-> Glyphs.app diffs.

qa/shaping/armenian_case.json
madig: Ugh, I keep stumblning over these... Maybe I should turn this script into a standalone tool, also taking into consideration the upstreamed fontbakery code.

Yeah, this needs a bit more thinking. I like the font name version idea but I use git worktrees for all the ongoing PRs and font-v kept blowing up in those scenarios, so I typically disabled it in the Makefile. Haven't looked at font-v recently though. Just overwriting on execution is a bit meh... I mean, yeah, but then we get a diff for every file every time we update things. We'd actually like to get less diffs :/

chrissimpkins:
> > Can you describe how you arrived at the edits in [c7e4890](https://github.com/googlefonts/googlesans/commit/c7e489093640193cadd8f9fe6068f9b5d6408df9)? I see the following comment in your commit:
> 
> > 
> 
> > > Using Glyphs 2.6.6.
> 
> 
> 
> I copy-pasted the names of all glyphs into a text file (for uprights and italics), one glyph per line, ssh-copied it over to a Mac and ran the following script inside Glpyhs.app:
> 
> 
> 
> ```python
> 
> import pprint
> 
> import logging
> 
> 
> 
> with open("/tmp/glyphlist_italic.txt") as f:
> 
>     names = [n.strip() for n in f.readlines() if n]
> 
> 
> 
> production_names = {}
> 
> 
> 
> for name in names:
> 
>     i = Glyphs.glyphInfoForName(name)
> 
>     if i is None:
> 
>     	logging.warning("No info for %s", name)
> 
>         continue
> 
>     if i.productionName is not None and name != i.productionName:
> 
>         production_names[name] = i.productionName
> 
> 
> 
> pprint.pprint(production_names)
> 
> ```
> 
> 
> 
> I then copy-paste the results out and insert them into a Python session and do 
> 
> 
> 
> ```
> 
> In [1]: for p in Path("source/GoogleSans/").glob("*.ufo"):
> 
>    ...:     if "Italic" in p.name:
> 
>    ...:         continue
> 
>    ...:     u = Font.open(p)
> 
>    ...:     u.lib["public.postscriptNames"].update(up)
> 
>    ...:     u.save()
> 
> 
> 
> In [2]: for p in Path("source/GoogleSans/").glob("*Italic*.ufo"):
> 
>    ...:     u = Font.open(p)
> 
>    ...:     u.lib["public.postscriptNames"].update(it)
> 
>    ...:     u.save()
> 
> ```
> 
> 
> 
> Isn't it wonderful.

ugh.

qa/shaping/armenian_case.json
chrissimpkins: Do you update every JSON file every time you change any of them, or add a new one?

chrissimpkins:
> I copy-pasted the names of all glyphs into a text file (for uprights and italics), one glyph per line, ssh-copied it over to a Mac and ran the following script inside Glpyhs.app:

So what will the process be going forward?  You intend to do this periodically, at every new import?

chrissimpkins:
> > I copy-pasted the names of all glyphs into a text file (for uprights and italics), one glyph per line, ssh-copied it over to a Mac and ran the following script inside Glpyhs.app:
> 
> So what will the process be going forward? You intend to do this periodically, at every new import?


And have you addressed Cosimo's request in https://github.com/googlefonts/googlesans/issues/419#issuecomment-986709109?

chrissimpkins:
Most of the diffs in the PS names are name changes.  But it looks like some of the PostScript names are *new* rather than changed.

e.g., the following in opsz17-wght380-GRAD-50 master (sorry, I can't even open the file diffs in a browser over the number of changes here to review in the GH UI, reviewing locally in git).

```xml
<key>hyphentwo</key>
<string>uni2010</string>
```

This is to be expected?  I was under the impression that we are only changing existing names to what Glyphs.app defines?  Why do we get new names?

madig:
> So what will the process be going forward? You intend to do this periodically, at every new import?

Yes, I have to.

> And have you addressed Cosimo's request in #419 (comment)?

Oops, will do.

> This is to be expected? 

Yes, because maybe not all glyphs got a proper name, and this is a sweeping redo-everything change.

chrissimpkins:
I don't understand the Glyphs naming convention.

There are some formatted like this:

```
uni0915094D0937
```

and others formatted like this:

```
uni0E0D_uni005F
```

(changed from previous `uni0E0D005F`)...

Commentary only.  Nothing to do.

chrissimpkins:
LGTM. The only outstanding issue is to post the issue report about prod glyph names to the glyphsLib repo. I removed the "Closes #419" on the OP.  We'll leave that issue open to track it.  That is where Cosimo posted the original request.  

Otw gtg here.  I'll merge this.

chrissimpkins:
Changelog updated in https://github.com/googlefonts/googlesans/commit/84e9a5aadf841b22ccf2f838d5be5593d4bb9d67

chrissimpkins:
Thanks Nikolaus!

madig:
> I don't understand the Glyphs naming convention.
> 
> There are some formatted like this:
> 
> ```
> uni0915094D0937
> ```
> 
> and others formatted like this:
> 
> ```
> uni0E0D_uni005F
> ```
> 
> (changed from previous `uni0E0D005F`)...

If your uni-ligature-names get long, you shorten them to just `uniAAAABBBBCCCC` etc. to stay in some length limit. Not sure about the last example though, but have learned not to question Georg on this.

PR #460, opened 2022-02-25
Add fontmake transformed components decompose filter

chrissimpkins:
Closes #459

Issue #458, opened 2022-02-25
Dotted circle attachment


madig:
We now fail on the new fontbakery check:

```
 >> Check:   com.google.fonts/check/dotted_circle
    Desc:    Ensure dotted circle glyph is present and can attach marks.
    Files:   GoogleSansText-Regular.ttf

     FAIL The following glyphs could not be attached to the dotted circle
          glyph: - uni0A75 - uni0A51 - And uni0A4D [code:
          unattached-dotted-circle-marks]
    Result: FAIL
```

Gurmukhi in this case.

chrissimpkins:
@kalapi see https://github.com/googlefonts/googlesans/issues/458#issue-1150389218

chrissimpkins:
Will work around this issue with #461 and I'll speak with Kalapi about it next week.

chrissimpkins:
Maybe we use this? https://github.com/googlefonts/ufo2ft/pull/593

chrissimpkins:
I spoke with Kalapi about this today.  He is going to look into the issue and will be back in touch with us.

madig:
The circle filter is very interesting! I think we should play with it once it lands (as we need to update ufo2ft then).

chrissimpkins:
@kalapi Will you please let us know if this is something that you can fit in this month?

chrissimpkins:
@madig I spoke with Kalapi today. He will push these updates tomorrow.

PR #461, opened 2022-02-25
Deactivate dotted circle fontbakery checks temporarily

chrissimpkins:
Temporarily works around #458 so that our CI will pass.  We will need to include the changes in #460 to receive a green light in the CI.


chrissimpkins:
Intentionally merging this with a failing fontbakery CI check so that we can rebase #460 on these edits to make the CI pass.

PR #463, opened 2022-03-03
Update filter list in normalization

madig:
I forgot this in https://github.com/googlefonts/googlesans/pull/460


chrissimpkins:
This updates the UFO writes with the new ufo2ft filter at script import time?

madig:
This keeps the filter list static on each import and normalization run.

Issue #464, opened 2022-03-08
Significant compile time performance hit with Ethiopic script import


chrissimpkins:
Seen in https://github.com/googlefonts/googlesans/pull/451

Our VF format compile times increased from ~ 6 - 8 mins to ~ 14 - 15 mins.  Let's review and discuss.  Something seems odd about the two fold increase in compile time with these changes.

Related: #298 
Related topic (undocumented on this repo): incremental transition to Rust based libs in fontmake / Rust based compiler development.

belluzj:
It could be due to the GPOS file size optimization:
* Ethiopic has lots of new GPOS to go through
* the GPOS optimization is still run in a silly way in fontmake (1 time per master, then the masters are merged with varLib, then once more on the result of the merge) when actually only the last time on the VF is needed. I was waiting for the fontTools config module to sort this out.

I could make a branch without the GPOS optimization just to see whether that's where the extra time comes from.

madig:
Note: on my PC, in `build-bangla`, turning on GPOS compression makes total compile times go from 8:30 mins to ~12 mins.

FWIW, e.g. in the VF, we waste ~half of the time on compressing intermediate masters that only need to be compressed in a final step:

![image](https://user-images.githubusercontent.com/380829/157433241-53cc37e1-9abe-47e8-b9cb-d7769a1a48d8.png)

belluzj:
Maybe I should propose a temporary solution in fontmake to turn off the compression for intermediate masters using the environment variable instead of waiting for the proper fonttools config module?

madig:
Mh yeah, maybe temporarily pop `FONTTOOLS_GPOS_COMPACT_MODE` from `os.environ` while doing the intermediates and putting it back for thefinal step? I think this needs some fiddling in fontmake though to figure out what the end goal is, as in, don't do this if you want to output interpolatable-ttf. Or something.

chrissimpkins:
Is there time to work on this soon'ish?  The commit-based CI is going to slow down significantly during the iterative work that happens during script imports.

madig:
(Anecdote: I keep removing the env var from the Makefile while iterating and then keep accidentally commiting the change. Grrr)

belluzj:
> Is there time to work on this soon'ish?

Yes we'll have a look, it's supposed to be relatively easy.

madig:
[...]

PR #468, opened 2022-03-17
Only build variable fonts by default

madig:
See https://github.com/googlefonts/googlesans/issues/464#issuecomment-1069647180.

I removed the upload-artifacts.yml job because I was unsure whether it would work if static and variable jobs are uncoupled and the jobs themselves do not put up their artifacts. So, now you need to get artifacts from the jobs directly.


madig:
Oh wait! I misread your comment, you want to stop _using_ `build-*` branches but otherwise leave them in. Hm. Then we still need to remove the upload job because the VF job does not put up artifacts?

chrissimpkins:
Yes! We will need to be able to build and QA the static files.  They are still used in production distributions.  We haven't moved to fully VF distributions (yet!)

chrissimpkins:
And we need the upload artifacts workflow.  I use the GHA repository builds for my own QA work.  And these are what we share with design teams as the final post-import WIP builds that use our build workflow.  I'd prefer to keep it this way rather than introduce an individual's local venv and build workflow definition in the process. (i.e., with a push of font files to a PR or IR thread)

chrissimpkins:
My suggestion did not involve any source modifications.  Only branch definition approach mods.  For future imports, we use a PR branch that is not prefixed with `build-*`.  It will only build and QA the VF's in the GHA runner environment.  It will not upload any artifacts for download (_FWIW, the build-* branches were originally intended for intermittent pushes to gain access to the files for human review_). We can push commits to a build-* branch when we are ready to include static instance tests and compile fonts for testers.  This branch doesn't need to be linked to the PR.  From a workflow standpoint, I will always do this once I dive into the review process after the merge happened and before the PR is merged to main.  You *can* do it if you want to see test results as you merge, iterate through issues that arise with design teams during the merge.  You'll need to do it to share the GHA compiled VF files with design teams for review.  And just prior to a tester push seems like a good point to include the instance tests.  We will want to avoid multiple rounds of review if we understand that there is a static font issue and we need to change the sources.

Sgty?

madig:
Ok! :)

chrissimpkins:
See https://github.com/googlefonts/googlesans/issues/464#issuecomment-1074357539

madig:
Noted.

Issue #356, opened 2021-07-19
Add Telugu script support


chrissimpkins:
The Telugu import will take place during Q3 2021 and we intend to release this work in the v6.000 build.

chrissimpkins:
We will tentatively begin the merge work in late August.  Confirming dates with all teams involved.

chrissimpkins:
Import will begin on Aug 27.  Universal Thirst confirmed that sources will be ready. DaMa team has required personnel lined up.

chrissimpkins:
Date revised to begin on Sept 3.  Kalapi will submit a PR by the end of next week.

chrissimpkins:
Moving to the v7.000 release.  The design team requires more time.  Working on a new import date.

chrissimpkins:
New import start date is Oct 7.   Sources/metadata/docs will be available for import by that day.

chrissimpkins:
Import was revised to Q1 2022 with release in v8.

I spoke with Kalapi today and he plans to have the sources in a PR by next week.

chrissimpkins:
Moved to v9 release

chrissimpkins:
@madig @belluzj Source delivery date confirmed as April 15.  We'll begin work on the contextual kerning issues at that stage.

cc @kalapi

kalapi:
Hello @madig, @belluzj & @chrissimpkins,

Attaching the latest package of the Telugu herewith. 

There may be a few changes over the course of the next few days.

Please let me know if you have any questions.

[2022-04-18_GoogleSansTelugu_V1.000_BETA01.zip](https://github.com/googlefonts/googlesans/files/8506667/2022-04-18_GoogleSansTelugu_V1.000_BETA01.zip)

madig:
Thanks. First import in #481, without the QA stuff. Even the reference fonts look broken across a span of the test HTML though? Our import looks even more broken (see below, longer strings = ours). What changes are you working on?

![telugu](https://user-images.githubusercontent.com/380829/164243290-abba398e-f256-419f-bfe3-f3647c4f4528.gif)

kalapi:
Hi @madig,

Can you send me this strings and identify the cluster that seems broken to you in the reference fonts?

madig:
Duh, wait, I was being stupid and forgot to compile the feature file.

madig:
[...]

PR #481, opened 2022-04-20
Import Telugu

madig:
See https://github.com/googlefonts/googlesans/issues/356.


chrissimpkins:
I spoke with Kalapi today and the changes in https://github.com/googlefonts/googlesans/pull/481/commits/68588249cacfd5871d010ecc01f7acaa47fc6b18 should address the contextual positioning issue. He's confirmed this in local builds with a different fontmake version (earlier v3 era fontmake) compiled from his local Glyphs sources.  It apparently works and he does not believe that we will need to make any changes to address his primary concern in Telugu.  We'll need to confirm that this is the case **following the source import with the pipeline definitions used in our repo builds** (will require us to push a commit to a build-* branch when ready).  There has been no testing performed on files compiled with our upstream definitions against production UFO sources as of the master-specific fea file changes that were pushed in revision https://github.com/googlefonts/googlesans/pull/481/commits/68588249cacfd5871d010ecc01f7acaa47fc6b18.  Kalapi will share a couple of text runs so that we can begin to review the issue.  We can test this by confirming that advance widths of the forms that he will document in this thread interpolate across the optical size and weight axes using Font Goggles.  The full set of text runs required to support our HB regression tests will follow in ~ one week.  Kalapi was encouraged by the findings on his local testing and he doesn't anticipate significant changes in advance of the Telugu design review.

chrissimpkins:
@madig Prefer to use new, imported fea paths rather than embedding the master-specific shaping source in the UFO master features.fea files? (https://github.com/googlefonts/googlesans/commit/68588249cacfd5871d010ecc01f7acaa47fc6b18)

chrissimpkins:
@madig I rebased this branch on the changes from #485 that just landed in main. There were no conflicts during the rebase.

madig:
What do you mean? I put the variating features into their own file to avoid duplication in the GRAD sources.

chrissimpkins:
> What do you mean? I put the variating features into their own file to avoid duplication in the GRAD sources.

I understand.  Just interested to know if you feel that complexity / maintenance effort is reduced by adding multiple new fea files with a multi-level import structure.  Sgtm!

madig:
Hm, I don't think so. I still have to dig into how to ensure the stuff lands in instances, haven't looked at that at all.

madig:
Alright. The VF now works as expected from the few checks I've done, but the instances have the kerning of the default source, so stuff is off in e.g. the Bold face. Need to investigate what fontmake is doing when compiling instances. I see a diff between the supplied Medium font and the VF Medium instance though:

telugu

This indicates that instance generation doesn't really work because there is only feature replacement going on, not interpolation like in the VF. To make it truly work, we probably have to mess with binary interpolation somehow or just cut instances from the VF.

madig:
@chrissimpkins I talked to Cosimo regarding instance generation, and he also names binary interpolation and instancing. I propose that we 

1. Get https://github.com/googlefonts/googlesans/pull/487 in
2. Change the instance build process to use `fonttools varLib.instancer` to cut static instances from the VFs
3. Work out any issues we find (leave composite overlaps as is to keep size down?)
4. Deal with Telugu

The instancer is in use by Google Fonts AFAIU and makes variating features straightforward to support.

chrissimpkins:
Sure, let's tackle #487 now and then begin to have a look at the new VF instancing to statics approach.  Thoughts about the binary diff that we are seeing with the pipeline updates in #487?

madig:
Todo: figure out how to get the new winAsc-/Descent into the statics... or do in recalc script?

chrissimpkins:
> Todo: figure out how to get the new winAsc-/Descent into the statics... or do in recalc script?

Hmm, they should be the same as the values in the VF.  Does the instancer change them?

madig:
Apparently not, or I'm missing something. I'll have to bundle up the recalc stuff in the script and then later port it to the instancer.

I changed the ascent/descent values in the last commit but the warnings were suspiciously absent from the VFs. Maybe it's really an instancer issue.

madig:
Oh, we remove the MVAR table from the VFs, so instancer has no chance of setting the correct values. Duh.

chrissimpkins:
So it chooses to change v metrics values that are provided in the original var font in the absence of instance specific data? That seems like a bug?

chrissimpkins:
I will look into the status of the mvar removal requirement before we go to significant trouble to address it in the pipeline or with something custom here.  This came from a fontbakery recommendation.

chrissimpkins:
Adding MVAR to the builds will address it?

madig:
I don't know if leaving MVAR in fixes everything, it's a hunch. I think MVAR is disabled because of some DirectWrite bug or something exposed in Firefox.

I can maybe rewrite the Makefile to remove the MVAR table in a separate step... but first I want to understand the win metrics issue better.

khaledhosny:
https://github.com/googlefonts/fontbakery/issues/3400

khaledhosny:
More info about MVAR in https://github.com/googlefonts/fontbakery/issues/3413

madig:
I vote in favor of leaving MVAR in.

So, I updated the win metrics to accomodate Telugu and the warnings went away I think.

chrissimpkins:
Thanks for the additional context Khaled.  Yes, agree.  It looks like the decision to include this as a fail in fontbakery was changed ~ Aug 2021 after the discussions in the above threads.  

OK to add MVAR table back if we need it to gen all valid static font metrics values with the new instance from VF format approach.  

@madig will you please let us know if adding MVAR back addresses the Win metrics issues in the static fonts?

madig:
I will add the MVAR back in the cut-instances PR. 

BTW, the metrics already are fixed now that I updated them for Telugu.

chrissimpkins:
Took a look last night.  It looks like we strip the table with the subsetter

madig:
Yes. All we have to do is not do that I think :)

chrissimpkins:
Should be a one-line change

https://github.com/googlefonts/googlesans/blob/c12b58bd4f7dd460f5a095c05474c8d2ed1885b0/scripts/internal/prune_font_binary.py#L45

madig:
I did that (and two more changes) in the #493 PR.

madig:
Green light from our checker :)

madig:
@chrissimpkins ready for review.

chrissimpkins:
Pushed `build-import-telugu` branch to run our static font format CI tests

chrissimpkins:
During static instance gen:

> subprocess.CalledProcessError: Command '['fonttools', 'varLib.instancer', '--quiet', '--remove-overlaps', '--update-name-table', '-o', '../build/GoogleSans/static/GoogleSans-BoldItalic.ttf', '../build/GoogleSans/variable/GoogleSans-Italic[GRAD,opsz,wght].ttf', 'opsz=18.0', 'wght=700.0', 'GRAD=0.0']' died with <Signals.SIGKILL: 9>.

madig:
Can't access the logs, but maybe memory pressure (sigkill)? Pypy will eat more, so we need to reduce parallelism or... use CPython instead. How manny cores does the runner get?

chrissimpkins:
@madig removed the build-import-telugu branch.  Should be ready to regen following rebase on #493

madig:
Here you go :)

madig:
Actually, fontbakery 0.8.9 complains about a bunch of new and erroring checks locally...

chrissimpkins:
> Actually, fontbakery 0.8.9 complains about a bunch of new and erroring checks locally... 

The nameID fails in the VF can be safely ignored.  Thoughts about the "Regular" and "Bold" instance axis values?  I'm not sure why these are not defined as documented in the rationale.

madig:
Uh, so, this seems to look for the "Regular" and "Bold" instance names in the italics by looking at the full instance name, but they are called "Italic" and "Bold Italic" there? This seems like a bug in fontbakery, unless I'm missing something? I'll open an issue.

madig:
I'm convinced that's a bug: https://github.com/googlefonts/fontbakery/issues/3818.

Disabling both checks.

chrissimpkins:
@kalapi Will you please have a look at the Telugu in the Roman and Italic variable fonts that we pushed in https://github.com/googlefonts/googlesans/suites/7163288797/artifacts/285303619 and confirm that these are production ready?

Ty!

chrissimpkins:
@kalapi Adding the static fonts to this request too given the new approach to slicing statics out of VF's to support the feature requirements in Telugu.  The statics can be found here:

https://github.com/googlefonts/googlesans/suites/7163288797/artifacts/285303618

Thanks again!

chrissimpkins:
D/W Kalapi today.  He reviewed the static and variable font files, and approved the Telugu for release as of commit https://github.com/googlefonts/googlesans/pull/481/commits/0d4d0b4c83fa9784e9a6f5494057e081619e6b86

chrissimpkins:
Note: we are adding master-specific Telugu feature files in this PR

chrissimpkins:
Pushed https://github.com/googlefonts/googlesans/pull/481/commits/d8521e6811b2919b88986fae1e173bd775e377c8 with Apache License v2 added to a new script that is added in this PR.  Otherwise LGTM and ready for merge.  Will wait for CI to pass and then merge.

chrissimpkins:
File size as of the Telugu script import:

Roman VF 3.2MB
Italic VF 3.4MB
Static instances ~1.5-1.6MB / instance

PR #485, opened 2022-04-27
Add stylistic set 09 with support for cap height scaled and centered symbols

### LINKED ISSUE #484: Support alternate copyright, registered trademark, and sound recording copyright (published) forms
Request from a product team:

"Is it possible to get an option to have symbols like the copyright one here to be proportional with the numbers?

Roboto worked that way"

Current default version:

image (1)

Desired optional alternate scale and baseline position:

image (2)

---

@chrissimpkins notes below:

We should address the registered trademark symbol with the same approach.

The glyph outlines are already available in the production sources as *.alt glyphs.  The feature support was removed following the decision from the creative oversight team to abandon alternate glyph forms. 

Per Bringhurst: "...the copyright symbol sometimes appears as a superscript, but its rightful place in typography is on the baseline."

We discussed whether to use the cap height scaled and centered forms as defaults and the decision was made to address the request in a new stylistic set instead. While this is maybe not typographically correct, copyright is used in lots of places across Google products and documentation, and a default change will lead to many layout diffs.  Rather than changing existing layout (and flagging on thousands of screenshot diff tests), we'll re-introduce as an optional alternate. 

This should involve a simple one-liner sub rule for each glyph in a new stylistic set.

chrissimpkins:
Closes #484

### DIFF
--- a/qa/check-fea.py
+++ b/qa/check-fea.py
@@ -95,6 +95,7 @@ STATIC_UPRIGHT_FEA = [
     "ss06",
     "ss07",
     "ss08",
+    "ss09",
     "subs",
     "sups",
     "tnum",
@@ -150,6 +151,7 @@ STATIC_ITALICS_FEA = [
     "ss06",
     "ss07",
     "ss08",
+    "ss09",
     "subs",
     "sups",
     "tnum",
@@ -206,6 +208,7 @@ VAR_UPRIGHT_FEA = [
     "ss06",
     "ss07",
     "ss08",
+    "ss09",
     "subs",
     "sups",
     "tnum",
@@ -261,6 +264,7 @@ VAR_ITALICS_FEA = [
     "ss06",
     "ss07",
     "ss08",
+    "ss09",
     "subs",
     "sups",
     "tnum",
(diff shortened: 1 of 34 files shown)


chrissimpkins:
Need to remove the glyphs from the skip exports list

chrissimpkins:
ufolib2 approach used to remove from lib in https://github.com/googlefonts/googlesans/pull/485/commits/2008c467429b465eab4e6521fe73406ad10d4354

```python
from ufoLib2 import Font
import pathlib

ROOT = pathlib.Path("[LOCAL PATH]")
PATHS = ROOT.glob("*.ufo")
REMOVE_LIST = ["copyright.alt", "registered.alt", "published.alt"]

for path in PATHS:
    font = Font(path)
    for glyph in REMOVE_LIST:
        font.lib["public.skipExportGlyphs"].remove(glyph)
        
    font.save(path, overwrite=True)
```

chrissimpkins:
and then I manually removed the glyph names from a duplicate set of skipExportGlyphs data in the designspace files... 🤦 

1e232f649881fb71e843cce93d6a7ae21b04933c

Evan Adams' two copies of the truth problem. :)

chrissimpkins:
Glyph set defs updated https://github.com/googlefonts/googlesans/pull/485/commits/10b6dc80b66652a1ebd83f19065b238638d65f93

ss09 fea tag added to CI fea tag checks https://github.com/googlefonts/googlesans/pull/485/commits/c896e4d05db9ca2a686e5c149612f568c6c8c361

chrissimpkins:
@madig I'd like to release this in v8.000.  How problematic will that be for the v9 targeted Telugu work going on in #481?

chrissimpkins:
The Telugu import branch cleanly rebases on this branch locally for me.  Let's merge.

Issue #483, opened 2022-04-21
v9 release compiler pipeline update issues


chrissimpkins:
There are a few open issues that will require an update in our production compile pipeline.  Adding an issue to track things that we come across in pipeline lib releases that might alter our builds and require review in the fonts/modification of the build process.

- Use harfbuzz GPOS/GSUB packer by default when uharfbuzz is present in the build environment as of fonttools v4.33.0 https://github.com/fonttools/fonttools/pull/2552 https://github.com/fonttools/fonttools/pull/2589

chrissimpkins:
There is a new major version release of the actions/setup-python (v3) GitHub Action that supports pip caching.  Will investigate a version bump to v3 and a transition to that pip dependency caching approach in our test and production builds.

PR #487, opened 2022-04-28
Update build dependencies for v9.000

### LINKED ISSUE #479: Presence of GPOS7 lookups
New fontbakery warning as of revisions + fontbakery release update that occurred in #474:

```
    Ensure no GSUB5/GPOS7 lookups are present.
    with build/GoogleSans/variable/GoogleSans[GRAD,opsz,wght].ttf

      Rationale:                                                                
      Versions of fonttools >=4.14.0 (19 August 2020) perform an optimisation on
      chained contextual lookups, expressing GSUB6 as GSUB5 and GPOS8 and GPOS7 
      where possible (when there are no suffixes/prefixes for all rules in the  
      lookup).                                                                  
                                                                                
      However, makeotf has never generated these lookup types and they are rare 
      in practice. Perhaps before of this, Mac's CoreText shaper does not       
      correctly interpret GPOS7, and certain versions of Windows do not         
      correctly interpret GSUB5, meaning that these lookups will be ignored by  
      the shaper, and fonts containing these lookups will have unintended       
      positioning and substitution errors.                                      
                                                                                
      To fix this warning, rebuild the font with a recent version of fonttools. 

     WARN Font contains a GPOS7 lookup which is not processed by macOS          
          [code: has-gpos7]                                                     

     WARN Font contains a GSUB5 lookup which is not processed by macOS          
          [code: has-gsub5]                                                     


    Result: WARN
```

Fix:

- bump our fontmake compiler pipeline to use a new version of the fontTools library dependency

chrissimpkins:
Related #483

Closes #479 

---

### TODO

- [x] review binary diffs in static fonts
- [x] review binary diffs in variable fonts


chrissimpkins:
Big diff

Here is the ttx diff dump for the Google Sans Roman VF built with defs in this branch vs. main. A quick glance suggests lots of GSUB/GPOS changes in the binary.

[googlesans-roman-vf.diff.zip](https://github.com/googlefonts/googlesans/files/8585180/googlesans-roman-vf.diff.zip)

chrissimpkins:
Addressed https://github.com/googlefonts/googlesans/issues/483#issuecomment-1112537401 in this branch

The new pip dependency caching approach appears to function without issues.

chrissimpkins:
Seeing this in the logged warnings:

> WARNING:fontTools.ttLib.tables.otBase:hb.repack failed to serialize 'GPOS', reverting to pure-python serializer; the error message was: RepackerError

madig:
This is expected: https://github.com/harfbuzz/harfbuzz/issues/3547

chrissimpkins:
> This is expected: [harfbuzz/harfbuzz#3547](https://github.com/harfbuzz/harfbuzz/issues/3547)

Expected means that we are running into repacking overflow issues so we rely on a fallback to the pure Python approach?

madig:
Yes, because hb-repacker only really works for the subsetting case right now it seems, not for the more general pack-anything case.

chrissimpkins:
Maybe we should toggle the attempt to use the hb packer off (if possible) or not install the uharfbuzz dep prior to build time then.  There is no need to use the compile time required to understand that the hb packing approach will fail.  And it might make sense to make the transition explicit if/when support is available rather than tease out that this happened with a future dependency update and review of binary diffs.

madig:
I see a lot of GPOS lookups type 7 turned into other types (8 but also others). Seems like https://github.com/fonttools/fonttools/pull/2540 is the most likely cause and intentional. Also, there's a gigantic section in the upright VF shuffling anchor data around, which may be caused by https://github.com/googlefonts/ufo2ft/releases/tag/v2.25.3.

Edit: using the old ufo2ft does not make the anchor stuff go away. It may be fonttools after all. I do see one conspicuous MarkBasePos table change where some Deva, Hebrew, Gurmukhi, Lao, Khmer and Bengali non-spacing marks newly appear and further down the MarkArray now has the same amount of entries as there are MarkBasePos entries. If you look at the previous MarkBasePos table, there were actually 2 parts to it with the second one containing the new glyphs above. It looks like the 2 subtables were simply merged.

chrissimpkins:
Reviewed by Nikolaus and OK to merge

PR #491, opened 2022-05-03
Use ufo2ft's default feature writers

madig:
I (re)learned that I shouldn't change the default feature writers and that `append` mode doesn't do what I thought it would: https://github.com/googlefonts/ufo2ft/issues/611.

We need this for Telugu, to make Kalapi's manually written blwm and dist features work as intended: they need to be moved to after the autogenerated features, which never happens in `append` mode. I also needed to change some existing feature files.

Going forward, use the Glyphs.app way of doing things: if you need to insert manually written feature code and mix it with autogenerated code, write the features and insert a `# Automatic Code` at the start of the block (meaning your code should follow after autogenerated code) or at the end (meaning it should be placed before autogenerated code).


chrissimpkins:
Required for Telugu import, likely future scripts.  D/W Nikolaus today.  Merged

PR #493, opened 2022-05-09
Cut instances from VFs

madig:
(Description updated with findings in latest state)

This PR cuts instances from the VFs instead of interpolating them from the sources, because we need variating features for Telugu.

There are huge amounts of binary diffs across the instances. The most interesting I saw:

* The `glyf` table has plenty of reordered contours and the overlap bit is set on a lot of glyphs. Harmless I suppose.
* The instancer will decompose overlapping composites, and the subsetter will subsequently remove a lot of glyphs that are only used for construction, so we drop 76 glyphs. This also leads to a lot of diffs in GPOS/GDEF.
* Quite a few of off-by-one diffs in `htmx` that shouldn't matter. Some off-by ones in ligature stuff, but that hopefully won't matter.
* Tiny off-by-one layout changes in some Armenian sequences.
* Instances will have a `STAT` table now and some additional `name` table entries
* Statics will be 100-200 KB larger, probably due to the overlap removal and less optimal GPOS compression time (after compile, instead of during).
* VFs will have an `MVAR` table.


madig:
I find in the cut instances:

* File sizes went from 1.4-1.5 MB to 1.6 MB.
* OS/2 PANOSE fields are zero
* OS/2 fsSelection bits 5 and 6 for regular/bold are reset
* Overlapping composites are decomposed
* Consequently, subsetting removes a lot of debris
* Some rounding changes here and there in the outlines themselves and the left side bearing value in the hmtx table
* Some contours reordered
* Countless changes in GDEF/GPOS
* The Text variants are now GoogleSans-TextMedium instead of GoogleSansText-Medium
* Shaping tests fails

chrissimpkins:
Ugh

chrissimpkins:
@nyshadhr9 Is it possible to have access to the Google Sans / Google Sans Text static instances that you build from the variable fonts in the web API pipeline so that we can compare with the approach that we are attempting here?  Ideally, the ones that are from our v8.000 release files that were pushed last week.

madig:
The regression tests (just two of them fail) come down to apparently off-by-one advance widths (spacing? kerning?) for some Armenian glyphs, most likely due to different rounding:

* GoogleSans-MediumItalic.ttf
  * uni055B.comb
  * uni055C.comb
  * uni055E.comb
* GoogleSansText-Medium.ttf
  * uni055B.short.comb
  * uni055C.short.comb
  * uni055E.short.comb

madig:
I copy-pasted some code from ufo2ft to re-set names for the text styles, but I seem to miss the mark. Continuing tomorrow.

madig:
Ok, I almost got the name table sorted and found that it, uh, is wrong for the Text variants. E.g. GoogleSansText-Medium.ttf has IDs 16 and 17 of "Google Sans Text" and "Medium" when it should be "Google Sans" and "Text Medium". Hrm.

chrissimpkins:
Google Sans Text has always been considered a separate family in the static builds.  That is intentional and it should remain that way.  The lineage is that this was a static first project that had two families: "Google Sans Display" and "Google Sans".  The GS Display family was deprecated around the time that we released GS Text and GS Text was added as a new family in the statics.  It was around the same time that we moved to the variable font format where the family became "Google Sans" and Text [Weight] were styles within the same family.  We should keep the previous approach.  Teams rely on that naming if they are using the static builds.

chrissimpkins:
On a positive note, Figma just released VF font format support this week. :)

madig:
Ok, I'll keep the naming the same. Or does it mean that we don't need statics anymore?

I'm currently trying to fix up the OS/2 and head rg/bd/it/bdit bits in a post-processing step. Are you concerned by the size increase? Or the rounding issues? Otherwise I'd proceed to base Telugu on this.

madig:
Bah. Make is fontmaking the VFs twice when making statics. I bet the targets are jumbled.

madig:
Additionally, fontbakery complains about metrics not checking out anymore.

E.g. GoogleSansText-Bold.ttf:

> The OS/2.usWinAscent value must be larger than the head.yMax value.  Received: OS/2.usWinAscent = 1320 head.yMax = 1322

> OS/2 xAvgCharWidth is 700 but it should be 744 which corresponds to the average of the widths of all glyphs in the font.

madig:
With this approach, Telugu statics seem to variate correctly.

madig:
@chrissimpkins my suggestion is that we keep running and take care of the details later, though I can foresee tiny screenshot diffs resulting from this.

chrissimpkins:
> @chrissimpkins my suggestion is that we keep running and take care of the details later, though I can foresee tiny screenshot diffs resulting from this.

What are the full set of diffs that we need to address at this stage? Everything is back to the original approach in name tables and style mapping (names/bits)? We are only dealing with some outline differences that are small off by one changes in Armenian script only, perhaps in other scripts that are not captured in our regression text runs?

madig:
There are various miscellaneous diffs across many tables, some inconsequential (e.g. some stuff in OS/2), others inscrutable (GPOS/GSUB, most likely due to more aggressive subsetting) and some relevant for screenshot diffs (many overlap removals may result in very subtle graphical diffs picked up by tools only). Plus those few Armn shaping diffs plus things not caught by the regression test suite, yes.

madig:
Todo: 
- [x] rewrite makefile to go over each instance font alone (simplify)
- [x] add GPOS compression to statics
- [x] one last pass over the diffs

madig:
@chrissimpkins I believe this is ready for consumption. I updated the description.

chrissimpkins:
Pushed `build-cut-instances` branch so that we run our static CI tests

madig:
Looks good. Except, wait a minute, are we double-building fonts in `static-artifact-ci.yml`, `variable-artifact-ci.yml` and `upload-artifacts.yml`?

chrissimpkins:
This looks great Nikolaus.  Thank you very much for all of your work on this transition.

chrissimpkins (approved):

PR #494, opened 2022-05-11
Add script to mark overlaps

madig:
As discussed, this is the script from https://github.com/google/fonts/issues/4405#issuecomment-1080664412.

This will probably mark more glyphs than strictly necessary. I need to figure out how to use Skia to test all paths for intersections.


m4rc1e:
Just a heads up but I haven't forgotten about this.

We're pushing Maven Pro, https://github.com/google/fonts/pull/4539 to the production server this week. This family has glyph overlap bits enabled using your script.

If we don't get any negative feedback, I'll finish my draft varLib implementation https://github.com/m4rc1e/fonttools/commit/f0c55451bc3ff030c670893cd6d6aa81bdd6308f. I'll then add some ufo2ft lib stuff as well, as you and Cosimo discussed.

chrissimpkins:
Thanks Marc!

chrissimpkins:
Looks like pathops needs to be added to our requirements.txt and requirements.in files

PR #496, opened 2022-05-17
Add generation of interpolatable masters

### LINKED ISSUE #471: Provide interpolatable TTFs as part of releases
For merging Google Sans into Google Sans Arabic (which will remain a separate family), we merge TTF masters and build variable fonts from them instead of doing a source merge with all the complications of differences in build tooling.

To do that, we need interpolatable TTFs for Google Sans. For now, I built them with manually with:
```
fontmake -m source/GoogleSans/GoogleSans.designspace -o ttf-interpolatable --output-dir build/GoogleSans/ttf-interpolatable
```

But it would be nice if there were a make target for this and the TTFs be part of the release (may be in a separate tarball), so that updating the Latin part of the Arabic font is as streamlined as possible.

(P.S. I tried instantiating masters from the variable font, and it worked for most part until I found out that quite a few glyphs ended up not being compatible after instantiation and varLib was ignoring them).

madig:
Add a Makefile target to generate interpolatable masters.

Should it be added to `all:`? 

Closes #471

### DIFF
--- a/source/Makefile
+++ b/source/Makefile
@@ -38,6 +38,7 @@ ITALIC_SOURCE=GoogleSans/GoogleSans-Italic.designspace
 FONT_BUILD_DIR=../build/GoogleSans
 STATIC_BUILD_DIR=$(FONT_BUILD_DIR)/static
 VARIABLE_BUILD_DIR=$(FONT_BUILD_DIR)/variable
+INTERPOLATABLE_BUILD_DIR=$(FONT_BUILD_DIR)/interpolatable
 
 STAGING_DIR=GoogleSans/staging
 
@@ -137,6 +138,14 @@ VARIABLE_BUILD_PATH_ITALIC_TARGET: $(FONT_BUILD_DIR)
 	font-v write --sha1 $(VARIABLE_BUILD_PATH_ITALIC)
 	mv $(VARIABLE_BUILD_PATH_ITALIC) "$(VARIABLE_BUILD_PATH_ITALIC_TARGET)"
 
+gs-compatible-masters: gs-compatible-masters-upright gs-compatible-masters-italic
+
+gs-compatible-masters-upright: $(FONT_BUILD_DIR)
+	$(FONTMAKE_EXE) -m $(UPRIGHT_SOURCE) -o ttf-interpolatable --output-dir $(INTERPOLATABLE_BUILD_DIR) --expand-features-to-instances
+
+gs-compatible-masters-italic: $(FONT_BUILD_DIR)
+	$(FONTMAKE_EXE) -m $(ITALIC_SOURCE) -o ttf-interpolatable --output-dir $(INTERPOLATABLE_BUILD_DIR) --expand-features-to-instances
+
 gs-vf-vendor: gs-vf-vendor-glyphs2designspace
 	$(foreach \
 		file, \
@@ -215,6 +224,7 @@ $(FONT_BUILD_DIR):
 	mkdir -p $(FONT_BUILD_DIR)
 	mkdir -p $(STATIC_BUILD_DIR)
 	mkdir -p $(VARIABLE_BUILD_DIR)
+	mkdir -p $(INTERPOLATABLE_BUILD_DIR)
 
 $(STAGING_DIR):
 	mkdir -p $(STAGING_DIR)
@@ -222,7 +232,8 @@ $(STAGING_DIR):
 .PHONY: gs-static gs-vf \
 gs-regular gs-italic gs-medium gs-medium-italic gs-bold gs-bold-italic \
 gst-regular gst-italic gst-medium gst-medium-italic gst-bold gst-bold-italic \
-gs-vf-upright gs-vf-italic gs-ufo2glyphs
+gs-vf-upright gs-vf-italic gs-ufo2glyphs gs-compatible-masters \
+gs-compatible-masters-upright gs-compatible-masters-italic
 
 # Disable built-in rules to speed up source globbing.
 MAKEFLAGS += --no-builtin-rules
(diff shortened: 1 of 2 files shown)


chrissimpkins:
> Should it be added to all:?

Does it need to be in the release archive?  If so, we should.  It looks like the GHA release workflow uses `make -j` to build the release.

chrissimpkins:
cc @khaledhosny re: #471 request

source/Makefile
```diff
@@ -137,6 +138,14 @@ VARIABLE_BUILD_PATH_ITALIC_TARGET: $(FONT_BUILD_DIR)
 	font-v write --sha1 $(VARIABLE_BUILD_PATH_ITALIC)
 	mv $(VARIABLE_BUILD_PATH_ITALIC) "$(VARIABLE_BUILD_PATH_ITALIC_TARGET)"
 
+gs-compatible-masters: gs-compatible-masters-upright gs-compatible-masters-italic
+
+gs-compatible-masters-upright: $(FONT_BUILD_DIR)
+	$(FONTMAKE_EXE) -m $(UPRIGHT_SOURCE) -o ttf-interpolatable --output-dir $(INTERPOLATABLE_BUILD_DIR) --expand-features-to-instances
```
khaledhosny: What is the purpose of `--expand-features-to-instances` here? (I have been building without it, so I hope I’m not missing something).

khaledhosny:
> Does it need to be in the release archive?

That would be ideal for me, so that there is a canonical source of these files.

source/Makefile
madig: Oh, this is primarily interesting for instance generation, I think I made a copy-paste oopsie.

chrissimpkins:
Sgtm.  Let's add it to the default all target and include in the release archives.  

@madig how will this work with the transition to statics instanced from vf’s targets?  

And should we add new CI workflows to cover the new build targets? This is a bigger question that influences our expected values for HB regression testing, expected glyph name/order, likey other QA

madig:
They are separate from all! If anything, I could base the VF builds off the interpolatable builds but will have to brush up on my Makefile-fu to figure that one out...

I don't think we need dedicated QA targets, as long as the command to generate them is the same as the one used to generate VFs.

chrissimpkins:
> If anything, I could base the VF builds off the interpolatable builds 

If this means a transition to a different approach to the vf format builds, let's not do that now.  We have a number of things to address and don't need another extensive binary diff review.

madig:
Aaaargh. I tried to base VF generation on ttf-interpolatable generation just to see what happens, and there are indeed binary diffs. A very cursory look suggests they are mainly reordering things but I can't be sure. Let's go with separate build steps then...

chrissimpkins:
We should land this and release a new v8.001 to support downstream family builds

chrissimpkins:
@khaledhosny We plan to push a new v8.001 release with the new build format this week.  Do things in these builds look ok for your use?

khaledhosny:
Is there a way to download the files built in this PR to double check locally? Otherwise the change looks good and should produce similar files to what I have been manually building.

madig:
We're not uploading artifacts from the build jobs themselves, so you'll have to run it locally...

khaledhosny:
I’m not sure if this is expected or not:
```
$ make gs-compatible-masters-upright
make: *** No rule to make target `gs-compatible-masters-upright'.  Stop.
$ make gs-compatible-masters-italic
make: *** No rule to make target `gs-compatible-masters-italic'.  Stop.
```

`make gs-compatible-masters` works, though. The files look fine, but they use production glyph names and the files I was building manually use source glyph names. I think source glyph names make more sense here since these are not the final production files.

chrissimpkins:
> I’m not sure if this is expected or not:
> 
> ```
> $ make gs-compatible-masters-upright
> make: *** No rule to make target `gs-compatible-masters-upright'.  Stop.
> $ make gs-compatible-masters-italic
> make: *** No rule to make target `gs-compatible-masters-italic'.  Stop.
> ```
> 
> `make gs-compatible-masters` works, though. The files look fine, but they use production glyph names and the files I was building manually use source glyph names. I think source glyph names make more sense here since these are not the final production files.

Thoughts @madig?

madig:
Added top-level targets and keep design names.

source/Makefile
```diff
@@ -38,6 +38,7 @@ ITALIC_SOURCE=GoogleSans/GoogleSans-Italic.designspace
 FONT_BUILD_DIR=../build/GoogleSans
 STATIC_BUILD_DIR=$(FONT_BUILD_DIR)/static
 VARIABLE_BUILD_DIR=$(FONT_BUILD_DIR)/variable
+INTERPOLATABLE_BUILD_DIR=$(FONT_BUILD_DIR)/interpolatable
```
chrissimpkins, line 41: For future reference, this PR will add a new set of ttf-interpolatable artifacts on the path `build/GoogleSans/interpolatable`.  We need these builds to support downstream G Fonts Latin Core glyph set merges into script-specific/style-specific GS families.

chrissimpkins (approved):

chrissimpkins:
Can confirm that our release.yml workflow uses `make -j` to compile the release archive.  We should include the new build format in all future releases.

chrissimpkins:
Running commit https://github.com/googlefonts/googlesans/pull/496/commits/917d1316d20d3276fad92974312d102726abedf1 on a build-* branch to execute the full suite of CI tests.

source/Makefile
chrissimpkins: These artifacts will be released in the git tagged GH Releases archives as of v8.001

chrissimpkins:
Merging so that we can release v8.001.

@madig is it helpful to include a workflow that will add these builds to downloadable archives that are pushed with commits on our build-* branches?  Or an extra branch / tag targeted build process that allows someone to receive the new format compiled from pre-release sources?

madig:
I just thought that maybe we can drop GPOS compression on interpolatable builds because why compress an intermediate artifact.

Hm... not sure. I think you typically want just the release package?

khaledhosny:
I think I tried with and without GPOS compression and saw no difference in the final Google Sans Arabic files.

khaledhosny:
(I mean building Google Sans masters w/o compression made no difference in my final Google Sans Arabic files)

PR #497, opened 2022-05-20
Draft new Makefile

madig:
I sketched a new Makefile that connects all the files properly, so you can run in parallel.

Todo: 

* [ ] The `**` wildcard to catch all files inside the UFOs doesn't quite work.
* [ ] Remove the MVAR table after the instances are cut, so the instancer has a chance to set correct values in various places.


madig:
Superseded by #493

PR #499, opened 2022-05-25
Add uharfbuzz to requirements.txt

### LINKED ISSUE #498: Fix release workflow
https://github.com/googlefonts/googlesans/pull/494 added a new uharfbuzz build requirement that is not included in our requirements.txt file.  We use requirements.txt for  release archive prep and the production release workflow now fails.

https://github.com/googlefonts/googlesans/actions/runs/2380995411

chrissimpkins:
Fixes #498

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,4 +1,3 @@
 -r requirements.txt
 
-uharfbuzz
 toml
--- a/requirements.in
+++ b/requirements.in
@@ -1,2 +1,3 @@
 fontmake[pathops]
 font-v
+uharfbuzz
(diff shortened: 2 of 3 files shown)


chrissimpkins:
The desire to pin dependency versions for reproducible builds and allow uharfbuzz versions to float for HB shaping regression tests are now in conflict after uharfbuzz was introduced as a build dependency in #494. Erring on the side to pin uharfbuzz version everywhere for now.  We'll need to untether the release artifact build process dependencies from the QA testing dev dependencies to get back to always current uharfbuzz for the HB shaping regression checks.

madig:
My intuition is to always pin uharfbuzz and have it be updated regularly.

PR #364, opened 2021-08-09
Use PyPy3

madig:
Supersedes https://github.com/googlefonts/googlesans/pull/299

PR #500, opened 2022-05-25
Use Pypy, take 3

madig:
Supersedes https://github.com/googlefonts/googlesans/pull/364.

Uses Pypy (in its own venv to avoid contamination) for building fonts.

* Using Pypy in 3.8 mode because of fontmake's forced dependency on lxml (fixed in main; I can't keep the current requirements _and_ stop pip from pulling in lxml). And there happen to be no lxml wheels for pypy3.9, so I have to use pypy3.8 instead. Probably just as fine but annoying.
* The responsibility for making and maintaining venvs goes into the Makefiles, which are now hardcoded to call into the created venvs. This means I can drop some dependency management from the GitHub Actions workflows. I think the fontbakery GHA action should be dropped as well in favor of just running tests from the Makefile, but maybe another day.


madig:
Judging from https://github.com/googlefonts/googlesans/actions/workflows/variable-artifact-ci.yml, this seems to reduce the VF job time by 1-5 mins, depending on variance, but we'll see...

madig:
Maybe we can merge this now 🤔 Upgrading to a newer fontmake would also require investigating diffs.

Makefile
```diff
@@ -17,6 +17,7 @@ STATIC_BUILD_DIR=$(FONT_BUILD_DIR)/static
 VARIABLE_BUILD_DIR=$(FONT_BUILD_DIR)/variable
 MASTER_UFO_DIR=$(FONT_BUILD_DIR)/master_ufo
 INSTANCE_UFO_DIR=$(FONT_BUILD_DIR)/instance_ufo
+PYTHON3=pypy3
```
chrissimpkins: Do we want to force users who perform local builds to install pypy3?  Does a pypy3 venv link the `bin/python` executable to the pypy3 interpereter so that we could just specify `python` in each of the commands and assume a user will install the Py/pypy version that they want to use in a local venv?  More flexible if so and doesn't break existing venv configurations based on python3

madig: * Hm... Well, I think "the users" are only us two?
* Yes, inside a venv, you can just call `python` and it will do the right thing.
* I think if you want to, you can do `make PYTHON3=python3.10 ...` and it will use that instead. Ideally you do that just for the setup step and all other commands activate the venv. I think we should use nox instead actually, like in the Loopless repo, because it takes care of this stuff. Some day.

madig:
@chrissimpkins maybe we can try a build-* branch from this?

This could land now and then I sort out the upgrade to a newer fontmake.

madig:
Interesting: turning on GPOS compression makes pypy compile slightly faster, maybe because it has to deal with fewer overflowing tables. 3:15 min vs. 3:26 min on my machine for compiling the upright VF with Telugu.

Sidenote: compressing while compiling, as is done with the env var, produces a ~20 KB smaller upright VF compared to compressing it after compilation :) So we want to keep doing that.

chrissimpkins:
This is turning on in the ttf-interpolatable build format with others already on? Or turning on across all builds vs not using it on any build?

madig:
It should use pypy for all the builds. I selectivly use pypy for workflow jobs that build fonts. Inside the Makefile, all should be done with the venv's Python, which I set to be Pypy by default.

madig:
I just looked into GPOS compression of stuff at various stages and how it impacts static file size, using this branch.

* Upright VF, uncompressed: 3951952 bytes
* Upright VF, compressed during compilation: 3209044 bytes
* Upright VF, compressed after compilation: 3229012 bytes
* Medium instance, cut from any of the three source VFs: 1804592 bytes
* Compressing it with otlLib.optimize: 1460488 bytes

So, cutting an instance requires that you optimize the tables afterwards, because apparently the compression of the source VF is undone. There is probably an opportunity to optimize the tables inside instancer before it moves to overflow resolution, potentially giving smaller files faster.

madig:
I tried to gauge the memory usage of Pypy, but the usual memory profilers don't seem to support Pypy... so I watched the Windows Task Manager during a `make -j` run (max 16 processes) instead. I got a roughly 10 GB peak RAM usage (first peak during font compilation, then even stronger when cutting instances).

madig:
@chrissimpkins I want to get this in, should I rebase it on main?

madig:
Edit: I just saw that fontbakery pulls in lxml. Hrm. It would be nice if Pypy is used just for building and normal CPython for tests. Nox could do that.

madig:
@chrissimpkins this is ready for a look methinks.

.github/workflows/build-incoming-files.yml
```diff
@@ -22,14 +22,9 @@ jobs:
...
           cache-dependency-path: "requirements*.txt"
-      - name: Install Python build dependencies
-        uses: py-actions/py-dependency-install@v2
-        with:
-          update-wheel: "true"
-          update-setuptools: "true"
       - name: Generate Glyphs.app file
         run: |
-          make gs-vf-vendor
+          make setup && make gs-vf-vendor
```
anthrotype: you're using `|` so you can have a multiline string, no need to `&&`

chrissimpkins:
Pushed `build-use-pypy3-v3` branch to run our static font format CI tests

chrissimpkins:
@madig

Can we remove the Generate Glyphs.app File from the build-* branch CI runs?  This failed with the move to pypy3, but it looks like it is because we don't use pypy3 in that workflow.

Makefile
chrissimpkins: Let's document how users can use their cPy3 install instead of pypy3 for compiles.  It looks like you landed on an approach that forces this by default with the Makefile targets

chrissimpkins:
@madig 

https://github.com/googlefonts/googlesans/pull/500#discussion_r899373401

madig:
Oh yeah, I should probably move those non-font-build related tasks to the top-level Makefile. Fix for now by using pypy3 for it.

chrissimpkins:
Can confirm that every main branch workflow (8) runs in parallel as of the latest commit https://github.com/googlefonts/googlesans/commit/f8967de2fbf1f6316e2123f8e5511f5ef4a49233

madig:
State: Chris didn't like how this mandates Pypy for building instead of making it a choice. Pypy should be isolated from CPython for testing because fontbakery pulls in lxml, which pessimizes build times. Need to keep lxml away from Pypy.

madig:
Making CPython and Pypy available at the same time seems tricky... argh.

madig:
So, Pypy shaves off 1-3 minutes 🤔

madig:
@chrissimpkins this now uses Pypy 3.x by default for building fonts, and to override that, you change the `PYTHON3_BUILD` env var.

I also changed the Makefile around to take over venv management, so you now don't need to activate the venv yourself.

.github/workflows/cicd.yml
```diff
@@ -57,15 +57,15 @@ jobs:
...
-          python-version: '3.x'
+          python-version: '3.10'
           cache: 'pip'
       - name: Build variable fonts
-        run: |
-          make setup
-          . .venv/bin/activate && make -j gs-vf
+        run: make setup && make -j gs-vf
+        env:
+          PYTHON3_BUILD: /opt/hostedtoolcache/PyPy/3.9.12/x64/bin/pypy3.9
```
anthrotype, line 68: why in actions/setup-python you set python-version to 3.10 but then here you call pypy with the full path? can't you tell github setup-python to use that?

anthrotype:
> Pypy shaves off 1-3 minutes

is it worth the potential extra risk of using an alternative python runtime (more bleeding edge, relatively less tested/widely used)?

madig:
I don't know, but I can't cross check against CPython 3.11 yet.

.github/workflows/cicd.yml
madig: I use python3.10 for the general venv and pyyp3 for the build venv.

anthrotype: why _not_ pypy for both?

madig: Because the general one needs deps for which there is no pypy wheel. babelfont (via glyphtools) e.g. requires orjson.

belluzj:
@chrissimpkins @MariannaPaszkowska Nikolaus is going to compare performance between the following scenarios:

|                                  | ufo2ft 2.29.0 (kern splitting off) | ufo2ft main (kern splitting on) |
|----------------------------------|------------------------------------|---------------------------------|
| Pypy 7.3.9                       |                        (compilation time in seconds; file size in kb)            |                                 |
| Pypy 7.3.9 with GPOS compaction  |                                    |                                 |
| Python 3.11                      |                                    |                                 |
| Python 3.11 with GPOS compaction |                                    |                                 |

Some results are already available below with Python 3.10, but Nikolaus predicts that with 3.11, pypy might lose some of its performance advantage, so maybe there will be no point in introducing it to the build system if we use 3.11. The results below are compilation times in seconds.

image

madig:
The newest data is:

grafik

Total time in seconds to fontmake the upright Google Sans VF, without and with kern splitting. Pypy remains faster when kern splitting is involved, which we hopefully ship soon in ufo2ft.

PR #501, opened 2022-05-25
Add designspace file cleanup following ttf-interpolatable builds

chrissimpkins:


chrissimpkins:
@madig the release archive includes two new designspace files in the `interpolatable` sub-dir.  Adding a step in the Makefile to remove them before release archive is prepped

chrissimpkins:
Check that.  The designspace files include the binary paths. It looks like that is how you use them?  Closing.

madig:
Nooo, keep them! We may need them.

PR #513, opened 2022-07-18
Remove stale glyphs

madig:
This PR removes glyphs that are marked to not be exported _and_ are unused by any other glyph.

There are some glyphs which use them, so the glyphs will be decomposed everywhere they're used, as before.


chrissimpkins:
There are now `housekeep_gha` <==> `remove_stale_glyphs` merge conflicts.

madig:
Updated approach and description.

PR #516, opened 2022-07-25
Update deps

madig:
Update dependencies.


chrissimpkins:
Need to assess binary diff vs. GS v9.000 release files

chrissimpkins:
@madig Will you please document the rationale for the dependency update?  What do you need for active work in progress?

madig:
Mainly, this is for getting Designspace v5 support, the uharfbuzz update in addition is to prevent an infinite loop during compilation. The ufo2ft update I'm after would contain https://github.com/googlefonts/ufo2ft/pull/636, which would roughly halve compilation times.

chrissimpkins:
> Mainly, this is for getting Designspace v5 support, the uharfbuzz update in addition is to prevent an infinite loop during compilation. The ufo2ft update I'm after would contain [googlefonts/ufo2ft#636](https://github.com/googlefonts/ufo2ft/pull/636), which would roughly halve compilation times.

How do you suggest that we approach the binary diff?  What is your level of assurance that the pipeline updates you propose here do not lead to regressions that impact users?

madig:
If we want the ufo2ft update (2.29.0b1+), then binary diffs will go multiple times around the block. The best way forward would be, I think, to expand the regression test suite (Kalapi provided large test pages with lots of (exhaustive?) combinations, more than what we put into the regression suite) plus fuzzing. I still need to look into Simon's SIESTA.

chrissimpkins:
Will expand our regression test suite and expand dependencies to bring in several other new performance-related updates.  Postponing.  Converted to Draft.

chrissimpkins:
Want to rebase on main and bump to current dependency versions, then we can begin to inspect with diffenator2. No rush. When there is time to do it and maybe before we dive in on new script imports.

cc @m4rc1e

madig:
Diffenator in progress: https://github.com/googlefonts/googlesans/actions/runs/3411212555

madig:
Hm. Diffenator says "No wordlist for Gurmukhi, Gujarati, Oriya, Tamil, Telugu, Sinhala, Lao, Ethiopic". That limits its regression testing scope.

madig:
(Display) Bold is showing two pixel diffs:

grafik

They are so miniscule that I had trouble finding them with a pixel magnifier.

madig:
@m4rc1e your opinion? :)

m4rc1e:
> Hm. Diffenator says "No wordlist for Gurmukhi, Gujarati, Oriya, Tamil, Telugu, Sinhala, Lao, Ethiopic". That limits its regression testing scope.

I plan to add these this week :-)

> @m4rc1e your opinion? :)

I am not Hermann Zapf so I don't see anything.

madig:
Let's rerun diffenator then :)

> I am not Hermann Zapf so I don't see anything.

But basically if the diff HTML stuff is empty then nothing changed?

m4rc1e:
I've added the missing wordlists so I'll rerun the ci now.

m4rc1e:
Results are in (took ~2 hours to complete all scripts 👎). Diffenator2 has flagged the Bold Telugu. On inspection the words look fine. 

I think I'm going to include the image pixels in the popups for each changed word/glyph. Atm, we only see the harfbuzz results. The image diff produces images which are 3pt (or em not sure) so we should get small image matrixes.

summary LGTM.

m4rc1e (approved):
LGTM.

I've added Text styles to the diff report, https://github.com/googlefonts/googlesans/actions/runs/3435413294

You can see that many words are flagged but nothing is perceptible to the human eye. 

I've now included the actual diff maps so when you hover over a glyph. It'll tell you what the diff for each pixel is (max val is 255).

Screenshot 2022-11-10 at 15 49 39

I spent some of today investigating whether my 3ppem FreeType bitmap approach is sufficient and it seems so. results can be found here, https://docs.google.com/document/d/16INOprdKWTZ4wyO41C0q4vuFpxFMO3Tod_Ig2sB4JAQ/edit

belluzj:
@chrissimpkins @MariannaPaszkowska The point of the PR is twofold:
1. update dependencies to have the latest and greatest of the fontmake pipeline: @madig can you please list the new things brought in by the dependency update and why we want them (faster, smaller fonts, keeping up with maintenance etc)
2. use this opportunity to check that diffenator2 can help assert that the fonts built with the new dependencies are visually the same despite different binary data (improve GPOS/GSUB compaction with Harfbuzz but no visual impact). Can this be confirmed from the tests above from @m4rc1e?
    * It's not 100% clear to me if the screenshot issues are real or artifacts of the tool. The fact that the differences are not perceptible to the human eye might not "be enough" since the Android screenshot tests are not human eyes and so might pick them up too? What do we think here? Is there a chance that diffenator2 is more picky than the Android screenshot tests?

madig:
The main updates are 

* fonttools (from 4.33.3 to 4.38.0): among other things, support Designspace v5
* fontmake (from 3.3.0 to 3.5.0): among other things, support Designspace v5 and not requiring lxml for Pypy (makes it easier to use)
* ufo2ft (from 2.27.0 to 2.29.0): among other things, support Designspace v5
* uharfbuzz (from 0.26.0 to 0.32.0): among other things, makes repacking GS work, resulting in slightly optimized GPOS tables.

There are dozens of smaller changes, bugfixes, etc. that we should keep up with.

m4rc1e:
> use this opportunity to check that diffenator2 can help assert that the fonts built with the new dependencies are visually the same despite different binary data (improve GPOS/GSUB compaction with Harfbuzz but no visual impact). Can this be confirmed from the tests above from @m4rc1e?

Correct. However, we can only assume the words we're testing are visually the same. There may be some combinations that we need to include in another wordlist to have full confidence e.g the google ligatures etc.


> It's not 100% clear to me if the screenshot issues are real or artifacts of the tool. The fact that the differences are not perceptible to the human eye might not "be enough" since the Android screenshot tests are not human eyes and so might pick them up too? What do we think here? Is there a chance that diffenator2 is more picky than the Android screenshot tests?

I reckon the Android diffs will also flag these errors. Due to the nature and high usage of Android, I reckon any changed pixel will get flagged. However, I reckon they'll be more willing to approve if we can mention what has changed to them before they catch it and that it's a none issue.

madig:
@MariannaPaszkowska I updated the deps here again and switched CI/CD to Python 3.11 for fastness.

@m4rc1e does diffenator2 support all the scripts/languages we have in GS so we can use it to check it that we don't regress?

madig:
Looking at the per-style word diff results (spot checking a few), I found this:

* Lots of supposedly changed text, but switching between old and new shows no perceptible diffs. Maybe 1 bit rendering diffs?
* Italic variants: Telugu seems to have changed kerning for e.g. నిర్దాక్షణ్యముగానుా. Need to investigate.

@m4rc1e anything to add?

madig:
Looking at crowbar, the new Telugu actually has a `dist` feature applied at the start of GPOS that wasn't applied before. Don't understand this yet.

madig:
So, the sources contain a kerning `@kern1.TEL_rrVocalic-telugu_L @kern2.TEL_k_ssa-telugu-R 20` and it wasn't applied before, but is now with updated deps. Diffing the debug feature file shows no change relating to Telugu, so I assume it's just more correct now.

@MariannaPaszkowska @chrissimpkins

m4rc1e:
@madig just taken a look at the html diff docs and yep, they don't show any changes. However, if I look at the raw BlackRenderer bitmaps which we use to work out what's different, there's clearly something happening.

Screenshot 2023-05-09 at 16 20 09

I'll check if this is an issue with blackrenderer or if a Linux based browser will show it as well.

Sorry in advance if my prose is terrible. I haven't slept much this week.

chrissimpkins:
Postponed to v12 so that we can land the Kannada and Malayalam before we introduce pixel level screenshot diffs that are problematic for rollout to production.

chrissimpkins:
Note: this branch is required for #574 to avoid a 3MB increase in overall file size.  We will need to land it in order to land the Gurage expansion.

cc @NeilSureshPatel @MarkJamra

m4rc1e:
I'll try and figure out what's happening today. If I use hb-view, the diffs are visible

Screenshot 2023-05-22 at 09 20 12

string: "ଆପଣ Android"

m4rc1e:
Apologies for the late response.

So I've checked Ubuntu FireFox and it also shows no differences. However, you previously said

> Looking at crowbar, the new Telugu actually has a dist feature applied at the start of GPOS that wasn't applied before. Don't understand this yet.

This is what's causing the issue. This dist feature is on by default when we pixel diff in blackrenderer. If I turn it off, we get the same results. I could disable this but there may be applications which have it on by default (haven't found any yet).

I recommend seeking out why a dist feature was added and discarding it if we don't need it. I see it's being used to make the space glyph wider for several writing systems. @kalapi is my assumption here correct?

cc @MariannaPaszkowska @madig

madig:
But how can a dist feature for Telugu influence all the other scripts?

m4rc1e:
Looks like kerning is also having an impact

```
hb-shape old-GoogleSans\[GRAD\,opsz\,wght\].ttf "Η Google"
[Eta=0+698|space=1+212|G=2+784|o=3+596|o=4+596|g=5+597|l=6+206|e=7+563]

hb-shape new-GoogleSans\[GRAD\,opsz\,wght\].ttf "Η Google"
[Eta=0+698|space=1+232|G=2+789|o=3+596|o=4+596|g=5+597|l=6+211|e=7+563]
```

If i disable the kerning:

```
hb-shape old-GoogleSans\[GRAD\,opsz\,wght\].ttf "Η Google" --features "-dist,-kern"
[Eta=0+698|space=1+232|G=2+789|o=3+596|o=4+596|g=5+597|l=6+211|e=7+563]

hb-shape new-GoogleSans\[GRAD\,opsz\,wght\].ttf "Η Google" --features "-dist,-kern"
[Eta=0+698|space=1+232|G=2+789|o=3+596|o=4+596|g=5+597|l=6+211|e=7+563]
```

We can see the space width is affected.

m4rc1e:
I'm wondering if splitting the kerns by script is causing this? I remember there was some talk about this months ago in some dependency which I've forgotten about.

madig:
Yes, but I think you're looking at at artefact. The kerning splitting uses the OpenType invariant that shaping should not cross script boundaries. HarfBuzz has a function to guess what's going on in a buffer, but does not do text run by script segmentation like a browser would do it, so when it sees a Greek character as the first one, it assumes the buffer is all Greek, meaning that with the split kerning, Latin is not getting looked at. In a browser, this wouldn't happen.

m4rc1e:
Ok good to know. I'll try and get diffenator2 to mimmic this behaviour. I have a few tasks on the todo pile so this may take me a while.

Since I've checked a range of platforms and browsers and found no diffs, I'm happy with this pr.

madig:
@MariannaPaszkowska diffenator2 needs to be changed to properly take the technical diffs from the update into consideration and avoid false positives. I discussed this issue with Marc, it has to do with text segmentation by script, and there are a few approaches to make it work, but they'd need some experimentation. Marc said he's added it to his todo list, I could also have a look, time permitting.

m4rc1e:
Diffenator2 update has been made. There shouldn't be any false positive reported now. Can someone rerun the diffenator2 action on this pr?

m4rc1e:
Thanks Nico! I've just updated the branch `diff-update-deps` so we should get new diffenator2 diffs.

madig:
@Hoolean's ufo2ft PR (https://github.com/googlefonts/ufo2ft/pull/739) caught the issue fixed by https://github.com/googlefonts/googlesans/pull/538 :) @MariannaPaszkowska can you please review and merge the latter?

madig:
@m4rc1e please force-update, had to rebase on main

m4rc1e:
Updated diffs can be found here: https://github.com/googlefonts/googlesans/actions/runs/5266444103

Diffs are much cleaner now that I'm not adding multilingual strings to a single harfbuzz buffer.

Only visual difference I can spot is for the string "(650) 555-1234". Kerning for "-1" has changed. "-3" and "-8" are also reported but I don't see a visual shift.

Screenshot 2023-06-14 at 12 47 54

madig:
I updated the dependencies again and ran diffenator2 with the AOSP strings corpus (which contains Kannada, which base diffenator doesn't). We found kerning difference in the following strings from Android:

* DVR ಮೂಲಕ ಪ್ರೋಗ್ರಾಂಗಳನ್ನು ರೆಕಾರ್ಡ್ ಮಾಡಲು ನಿಮಗೆ ಸಾಧ್ಯವಾಗುತ್ತದೆ. ಅದಾಗ್ಯೂ DVR ಗೆ ಕೆಲಸ ಮಾಡಲು ಇದೀಗ ನಿಮ್ಮ ಸಾಧನದಲ್ಲಿ ಸಾಕಷ್ಟು ಸ್ಥಳಾವಕಾಶವಿಲ್ಲ. ದಯವಿಟ್ಟು GB ಅಥವಾ ಅದಕ್ಕಿಂತ ಹೆಚ್ಚಿನ ಬಾಹ್ಯ ಡ್ರೈವ್‌ಗೆ ಸಂಪರ್ಕಿಸಿ ಮತ್ತು ಸಾಧನ ಸಂಗ್ರಹಣೆಯಂತೆ ಫಾರ್ಮ್ಯಾಟ್‌ ಮಾಡಲು ಹಂತಗಳನ್ನು ಅನುಸರಿಸಿ.
* ಒಂದಕ್ಕಿಂತ ಹೆಚ್ಚು ಬಾರಿ ಸ್ಲೈಡ್‌ಶೋ ಪ್ಲೇ ಮಾಡಿ
* 1 ದಿನಕ್ಕಿಂತಲೂ ಕಡಿಮೆ ಉಳಿದಿದೆ
* ಕಠ್ಮಂಡು
* ಮೌಂಟೇನ್ ವ್ಯೂ • H: --° L: --°
* ಕೌಲಾಲಂಪುರ್
* DVR ಮೂಲಕ ಪ್ರೋಗ್ರಾಂಗಳನ್ನು ರೆಕಾರ್ಡ್ ಮಾಡಲು ನಿಮಗೆ ಸಾಧ್ಯವಾಗುತ್ತದೆ. ಅದಾಗ್ಯೂ DVR ಗೆ ಕೆಲಸ ಮಾಡಲು ಇದೀಗ ನಿಮ್ಮ ಸಾಧನದಲ್ಲಿ ಸಾಕಷ್ಟು ಸ್ಥಳಾವಕಾಶವಿಲ್ಲ. ದಯವಿಟ್ಟು GB ಅಥವಾ ಅದಕ್ಕಿಂತ ಹೆಚ್ಚಿನ ಬಾಹ್ಯ ಡ್ರೈವ್‌ಗೆ ಸಂಪರ್ಕಿಸಿ ಮತ್ತು ಸಾಧನ ಸಂಗ್ರಹಣೆಯಂತೆ ಫಾರ್ಮ್ಯಾಟ್‌ ಮಾಡಲು ಹಂತಗಳನ್ನು ಅನುಸರಿಸಿ.
* ಭೂಕಂಪ ಮತ್ತು ಸುನಾಮಿ ಮುನ್ನೆಚ್ಚರಿಕೆ
* ಸರ್ವರ್‍ನಿಂದ ಇಮೇಲ್ ಅಳಿಸಿ
* ಅತಿಥಿ ಪ್ರೊಫೈಲ್‌ಗಳಿಗೆ ಕಂಪ್ಯಾನಿಯನ್ ಸಾಧನವನ್ನು ಪ್ರವೇಶಿಸಲು ಸಾಧ್ಯವಿಲ್ಲ
* ಓವರ್‌ಡ್ರಾ ಕೌಂಟರ್‌ ತೋರಿಸು
* ಪೂರ್ಣಗೊಳಿಸಲು ಇನ್ನೊಂದು ಸಾಧನವನ್ನು ಟ್ಯಾಪ್ ಮಾಡಿ
* MMS ಸಂದೇಶಗಳನ್ನು ಸ್ವಯಂಚಾಲಿತವಾಗಿ ಹಿಂಪಡೆದುಕೊಳ್ಳಿ

Plus the known one:

* (650) 555-1234

We narrowed down the diffs to the anusvara-kannada, here is the old font:

image

And here the new one:

image

Close-up:

image-1
image-2


Judging from the code in ufo2ft, this is an intentional change, where the anusvara-kannada is considered a spacing mark and therefore stops kerning between ka-kannada and pa-kannada while other non-spacing marks would be ignored.

@kalapi Does this reflect your intentions?

MariannaPaszkowska:
Thank you everyone for your contributions and testing of this PR! 🚀

PR #522, opened 2022-08-12
Configure GitHub Action runner infrastructure to 64 cores + 256GB RAM

chrissimpkins:


chrissimpkins:
@belluzj @madig @Hoolean @simoncozens I set up the configurable GH Actions runner environment. This should be plenty of processors and memory to continue our faster compile experiments...

PR #524, opened 2022-08-30
Automate figma workaround as part of the pipeline

### LINKED ISSUE #521: Google Sans VF test build instance definitions for Figma auto-opsz
**This is a test implementation and not currently intended for production release.**

Figma requires all fvar/STAT instance definitions to be defined at the optical size axis default value in order to toggle automatic VF optical sizing for text size on by default.  Let's build Roman and Italic test VFs that have the following instances, all defined at opsz 18:

- Regular
- Medium
- Bold
- Italic
- Medium Italic
- Bold Italic

The change removes all "Text" instances.  

We'll use this set of fonts to test the static to VF format transition in Figma.

Hoolean:
## Motivation

Closes #521.

## Progress

1. [X] Implement a script to automate ["fvar" table modification](https://github.com/googlefonts/googlesans/issues/521#issuecomment-1211839192).
2. [ ] Test that the automatically patched TTF functions correctly in Figma, looking out for regressions.
3. [ ] Create a Figma build target using the script.
4. [ ] Build the target in the pipeline.

### DIFF
--- a/.github/workflows/cicd.yml
+++ b/.github/workflows/cicd.yml
@@ -69,7 +69,9 @@ jobs:
       - uses: actions/upload-artifact@v3
         with:
           name: variable-fonts
-          path: 'build/GoogleSans/variable'
+          path: |
+            build/GoogleSans/variable
+            build/GoogleSans/.intermediate
 
   check-variable:
     needs:
@@ -86,7 +88,7 @@ jobs:
       - uses: actions/download-artifact@v3
         with:
           name: variable-fonts
-          path: 'build/GoogleSans/variable'
+          path: 'build/GoogleSans'
       - name: Font Bakery GS profile tests
         uses: f-actions/font-bakery@v3
         with:
@@ -135,7 +137,7 @@ jobs:
       - uses: actions/download-artifact@v3
         with:
           name: variable-fonts
-          path: 'build/GoogleSans/variable'
+          path: 'build/GoogleSans'
       - name: Build static fonts
         run: |
           make setup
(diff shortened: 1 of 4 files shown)


Hoolean:
Hi @chrissimpkins! As discussed, I am documenting the issue we encountered with Figma here.

It takes deliberate effort to trigger, and appears rare, having only occurred at sparse and unusual optical size values so far. It may also be a coincidence that we first noticed it with this PR's builds.

Should we pause the PR until we have confirmed this, or are we happy to continue?

---

## Bug

At certain positions on the optical size slider, 'Go, go, go!' is incorrectly rendered: every comma is substituted for 'Ǎ'.

### Reprod

1. Create a Figma document with the text 'Go, go, go!'.
2. Set the font to a patched upright Google Sans TTF (e.g. one of [Jany's original ones](https://github.com/googlefonts/googlesans/issues/521#issuecomment-1212028426)).
3. Set 'weight' to 400 and 'grad' to 0.
4. Wiggle the 'optical size' slider until a flash of 'Ǎ' is seen. The slider can also be released over this area to keep the artefact visible.

The issue could not be recreated by setting the value of 'optical size' to 17.11 manually. It only reliably occurred after repeatedly stopping and starting wiggling the axis for over a minute.

Furthermore, when a document with the artefact was saved, the issue appeared to resolve itself when the cached text was redrawn. This makes it difficult to capture in a Figma document.

### Environment

* Windows 10
* Desktop version of Figma
* Fonts:
  * Jany's manually patched TTFs
  * This PR's automatically patched TTFs

So far, the issue could not be recreated outside of Figma (e.g. in our font tester), and has not been tested with our unpatched fonts.

### Screenshot

Text rendered incorrectly in Figma

### Cause

Further testing is needed to know if our patched TTF is misbehaving, or if the changes are merely a coincidence; the lightness of the changes and the infrequency of the issue suggests the latter.

If the issue can be recreated with an unpatched Google Sans TTF, then the cause is not this PR; if the issue can be recreated with a different font, then the cause is not this project; in any case, it is likely that Figma's behaviour is incorrect.

PR #533, opened 2022-09-27
Add a normalization step for public.glyphOrder

### LINKED ISSUE #590: .notdef is not the first glyph in the glyphOrder
the ufos in source/ directory all contain public.glyphOrder key in their lib.plist, however the .notdef glyph is not listed as the first glyph, as it should be.
This is relying on the fact that ufo2ft will ignore this and automatically reorder the .notdef (if present) to be the first glyph in the font. This is required otherwise whatever happens to be the first glyph in the font with glyph index 0 will behave as if it has no mappings in the cmap table thus unaccessible.
currently fontc takes the provided pubic.glyphOrder at face value and applies it as is without any such reordering of the .notdef glyph, leading to invalid placement of the .notdef glyph.
We can definitely fix this in fontc compiler but I would also recommend that you fix this in the sources by moving .notdef glyph at the top of the public.glyphOrder, as it already is the case for the compiled fonts anyway.

belluzj:
I spotted while cleaning up the Latin extension branch that the public.glyphOrder had leftover entries that had been since deleted from the fonts. It's not an issue but maybe could be normalized. Also it seems some UFOs are missing slash.loclTAML.

Fixes #590 
Needs #516 first

### DIFF
--- a/scripts/internal/normalize.py
+++ b/scripts/internal/normalize.py
@@ -19,6 +19,7 @@ from pathlib import Path
 from typing import Dict, List, Set, Tuple, Union
 
 import ufo2ft
+from ufo2ft.util import makeOfficialGlyphOrder
 import ufoLib2
 from fontTools.designspaceLib import (
     DesignSpaceDocument,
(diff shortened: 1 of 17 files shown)


chrissimpkins:
Thanks Jany.  Marianna and I noticed the trailing localized Tamil slash at the end of the glyph order QA lists with new glyphs inserted in order before.  This is an indication that the glyph is missing from the source files?

belluzj:
No, the glyph is in the sources, see here: https://github.com/googlefonts/googlesans/blob/main/source/GoogleSans/GoogleSansItalic-opsz18-wght380-GRAD0.ufo/glyphs/slash.loclT_A_M_L_.glif

I think what went wrong is that when we imported it we forgot to add it to the public.glyphOrder, and so instead of being locked in one place, it was "trailing" at the end of the glyph set. I added it to the glyphOrder in the Latin Extension PR so that it doesn't move around anymore.

belluzj:
@MariannaPaszkowska @chrissimpkins I rebased this, it's ready to review and merge.

MariannaPaszkowska (approved):

PR #538, opened 2022-10-04
Remove "*strait*" glyphs

madig:
Older name used in Glyphs.app.

No binary diffs in the VFs, so I suppose they were dead weight.


madig:
@chrissimpkins @MariannaPaszkowska Ready.

belluzj:
This is a good candidate for merging just after the v10 release, as it doesn't impact the final font and is just a source cleanup.

MariannaPaszkowska:
LGTM, @chrissimpkins, can I merge this, or would you like to have a look too?

PR #539, opened 2022-10-05
Fix regression test metadata

### LINKED ISSUE #303: Harfbuzz shaping input files use ISO 15924 script tags
Documented @ https://en.wikipedia.org/wiki/ISO_15924

It looks like this is the problem with the Devanagari shaping files.  We are using Microsoft OpenType spec script tags in the shaping files and they are not always the same. The docs state:

> Script tags generally correspond to a Unicode script. However, the associations between them may not always be one-to-one, and the OpenType script tags are not guaranteed to be the same as Unicode Script property-value aliases or ISO 15924 script IDs.

The hb-shape help states:

>  --script=ISO-15924 tag                 Set text script (default: auto)

When I convert the `dev2` lang tag to `deva` hb-shape shows the expected localized output with `locl` text run testing.

cc @madig @kalapi

madig:
Closes https://github.com/googlefonts/googlesans/issues/303

* Correct script tags for Bangla and Telugu
* Remove useless DFLT and dflt script and language tags
* Correct a syntax error in one TOML file
* Create the input file for locl_latn.json, which went missing somehow.
* Write commit IDs into the input files because updating the output files means sifting through unwanted git changes every time.


madig:
@chrissimpkins @MariannaPaszkowska Ready.

chrissimpkins:
Can we re-gen the HB shaping output based on the v9.000 release commit so that we confirm no regressions associated with the post-release S Asian script changes targeting v10?

cc @MariannaPaszkowska

madig:
Regenerating the files with v9 confirms that nothing changes except Odia and Sinhala, which didn't exist at that point.

@chrissimpkins @MariannaPaszkowska I believe this is ready to go in.

PR #544, opened 2022-10-13
WIP: Diffenator2 integration 

m4rc1e:
Diffenator2 is still a wip but I think we can at least start testing it on this repo.

I'm going to keep adding commits so let's not review it yet.

First test. Can we generate proofing docs with linux screenshots.

### DIFF
--- a/.github/workflows/diffenate.yml
+++ b/.github/workflows/diffenate.yml
@@ -1,0 +1,88 @@
+name: Compare VFs against previous release
+
+on:
+  push:
+    branches:
+      - main # remove this since one workflow is working well. We only want to run this on diff-* branches.
+      - "diff-*"
+  # rm as well once this is working correctly
+  pull_request:
+
+jobs:
+  build-variable:
+    runs-on: [googlefonts-64cores-256GB]
+    steps:
+      - name: Check out source repository
+        uses: actions/checkout@v3
+      - name: Set up Python v3.x environment
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.x'
+          cache: 'pip'
+      - name: Build variable fonts
+        run: |
+          make setup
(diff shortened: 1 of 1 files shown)


m4rc1e:
I'll be fixing the failing action in https://github.com/m4rc1e/mavenproFont/pull/8 instead of hammering your ci.

m4rc1e:
Ok, screenshotting works. We get glorious 120 mega pixel size images. Next up diffing.

Images here: https://github.com/googlefonts/googlesans/suites/8753429767/artifacts/396700536

m4rc1e:
I'm opening up issues in https://github.com/m4rc1e/diffenator2/issues for anything which seems off.

Italics are not showing correctly in html.

m4rc1e:
Finally got diff screenshots working, https://github.com/googlefonts/googlesans/suites/8759156983/artifacts/398104874 (120mb download).

Next up running diffenator on the beefy linux machine.

m4rc1e:
Ok, diffenator is now working:

Screenshot 2022-10-14 at 11 33 38


It currently takes ~40 mins on the mega linux machine to produce all the diffs. I'll profile it to see where the bottlenecks are. I'm praying we can make some easy wins since I don't want to RIIT (Rewrite it in rust) atm.

A compromise for the time being is to only run diffenator when we've pushed to specific branches. I'll work on that after lunch.

@belluzj @chrissimpkins @MariannaPaszkowska the first version should be ready before next Friday. Happy to have a meeting about it and to discuss how to use it

m4rc1e:
@MariannaPaszkowska @chrissimpkins This is ready to merge so I can demo it for tomorrow.

The workflow produces some warnings that certain features are going to be deprecated soon. These issues are being caused by the actions this workflow uses, not the workflow itself. I will address these later on.

PR #553, opened 2022-11-08
Use Designspace v5 format

### LINKED ISSUE #551: Use Designspace v5
This would unify two Designspace files and make the separate STAT table script redundant.

1. Last time we tried this, we stumbled over the two Designspace files having different skipExportGlyph lists, which I think are the same for now. fontmake/ufo2ft still need to be changed to read lib keyes from a <variable-font> element I think?
2. We need updated requirements most likely, which is blocked on better regression testing.

madig:
This collects all variable info in one file.

The gs-stat.py script is now redundant.


madig:
Findings so far:

* The separate Designspaces called the opsz axis "Size" and the gs-stat.py script called it "Optical Size", leading to both names turning up in places in the font. The unified Designspace calls both the same. I suppose this is the right thing to happen.
* The name table is organized differently, as in, things are in different order, but other tables referencing name IDs stay the same, as do the important names in the statics. So I suppose this is not a problem.
* The name table is missing Mac names for some reason; the subsetter cuts them out. It's easy enough to make it stop doing that, though.
* The fvar table shows PostScript names now, which also treat Text and Display as one family. I filled in the PostScript names by hand in the DS, making Text a separate family.
* There is an annoyance described in https://github.com/fonttools/fonttools/issues/2877 and the workaround of adding `elidedfallback="Regular"` adds a redundant name entry to both fonts; which is "wrong" for the italic, but harmless, because the italic VF does not use it.

RickyDaMa:
Good news: diffenator2 reports no difference in outlines (phew)

[Diffenator2 report (v12.000 vs ds5).zip](https://github.com/user-attachments/files/15912884/Diffenator2.report.v12.000.vs.ds5.zip)

Observations from upright VF TTX diffs:
  - `caretSlopeRise` in `hhea` has changed from 1 to 1000
  - Lots of new `name` entries, I think caused by 7178e10
  - `fvar` instances have postscript names now (GoogleSans-Regular, GoogleSans-Medium, and GoogleSans-Bold)
  - `opsz` changed from "Size" to "Optical Size" in `fvar` (5811df0)

If you want to take a look at the TTFs yourself, they're included in the diffenator report archive

madig:
I've been looking at the name changes and believe there are no relevant ones. 

1. There are Mac-platform copies of the Windows-platform names now, probably because they aren't dropped anymore, as Ricky pointed out.
2. Various names have changed order
3. Some entries like "Grade" now are not duplicated anymore
4. There are now PostScript names for instances
5. The axis "Size" is now called "Optical Size"
6. The names for the Text-family instances are dropped because they aren't in fvar

source/GoogleSans/GoogleSans.designspace
```diff
@@ -1,89 +1,196 @@
...
+        <axis-subset name="Optical Size"/>
+        <axis-subset name="Weight"/>
+        <axis-subset name="Grade"/>
+        <axis-subset name="Italic" uservalue="1"/>
+      </axis-subsets>
+    </variable-font>
+  </variable-fonts>
   <instances>
-    <instance name="Google Sans Regular" familyname="Google Sans" stylename="Regular" filename="../../build/GoogleSans/instance_ufo/GoogleSans-Regular.ufo">
+    <instance name="Google Sans Regular" familyname="Google Sans" stylename="Regular" filename="../../build/GoogleSans/instance_ufo/GoogleSans-Regular.ufo" postscriptfontname="GoogleSans-Regular">
```
RickyDaMa, line 187: This (and ones like it) should be updated to just be "Regular"

RickyDaMa (changes requested):

madig:
I am inclined to recommend to just close this. We'd need more time to adjust import scripts to make this work, plus, Glyphs 2 doesn't support incompatible masters in the same file (and our glyphsLib is a bit old for Glyphs 3 support, so we'd need to update requirements and do regression testing, etc.). We don't gain much.

MariannaPaszkowska:
We are planning to move to fontc, and DSv5 is not yet supported there.

PR #554, opened 2022-11-09
diffenate.yml: add text styles

m4rc1e:
Diffenator currently only checks the Display styles. This PR will ensure we also check the Text styles.


m4rc1e:
@chrissimpkins this is good to merge.

PR #557, opened 2022-11-16
Use CPython 3.11

madig:
To compare against Pypy.

Compile times roughly 12 mins.


chrissimpkins:
I seem to recall that cPy 3.11 was reverted in GS Flex because of some dependency issues?  Where is the pipeline with cPy 3.11 support at this stage?

madig:
We're using 3.11 on Flex because we have new dependencies :)

Issue #558, opened 2022-11-16
Update to avoid deprecated functionality in GitHub workflows


Hoolean:
Our current workflows give deprecation warnings, and so we should review these and update our approach as necessary.

* [ ] ["Node.js 12 actions are deprecated"](https://github.com/googlefonts/googlesans/actions/runs/3432131167)
   * https://github.com/googlefonts/googlesans/pull/559
* [ ] ["The `set-output` command is deprecated and will be disabled soon"](https://github.com/googlefonts/googlesans/actions/runs/3266777549)

PR #559, opened 2022-11-16
Bump actions from the deprecated Node 12 to Node 16

Hoolean:
This PR bumps two actions to their next major version, to partially address #558. Between the old and new versions, neither action makes any behavioural changes (within the scope of our usage) aside from addressing the deprecation of Node 12:
- [f-actions/opentype-sanitizer](https://github.com/f-actions/opentype-sanitizer/blob/master/CHANGELOG.md#v200)
- [f-actions/fontbakery](https://github.com/f-actions/font-bakery/blob/master/CHANGELOG.md#v200)

Next, we can compare the QA CI output before and after the change to reassure ourselves of the versions' equivalence, otherwise we can proceed directly with review and merging if we are confident that the bumps have no other effect 🚀

### DIFF
--- a/.github/workflows/cicd.yml
+++ b/.github/workflows/cicd.yml
@@ -88,28 +88,28 @@ jobs:
           name: variable-fonts
           path: 'build/GoogleSans/variable'
       - name: Font Bakery GS profile tests
-        uses: f-actions/font-bakery@v1
+        uses: f-actions/font-bakery@v2
         with:
           version: 'latest'
           subcmd: 'check-profile'
           args: '--auto-jobs -C --loglevel WARN qa/check-googlesans.py'
           path: 'build/GoogleSans/variable/*.ttf'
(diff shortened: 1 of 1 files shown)


Hoolean:
(these actions are only used within the QA portion of our CI, in `check-variable` and `check-statics`, and so this change does not impact the final built fonts)

Hoolean:
It looks like this PR missed some usage of Node 12 in `release`, where both [actions/create-release](https://github.com/actions/create-release) and [actions/upload-release-asset](https://github.com/actions/upload-release-asset) are affected. Sadly, we cannot correct this with version bumps, as both actions are now archived; we will have to move to alternatives when support for Node 12 is dropped.

chrissimpkins (approved):

chrissimpkins:
Ty!

PR #564, opened 2022-12-14
Pin CI Python to 3.10

madig:
This unbreaks CI.

Before, we were requesting Python "3.x", which is a moving target, and we would need to update dependencies to make it work properly. Therefore, do the easy thing and request 3.10 instead, until we are ready.


madig:
@MariannaPaszkowska Ready to merge. Need this for the direct current PR.

PR #517, opened 2022-07-26
Latin extension 

elenaperaltad:
Todo:
- [x] Provide production names for new glyphs

PR #379, opened 2021-09-02
Fix `Eng` and `eng` tail outlines

chrissimpkins:
Fixes #284

Sources imported from David Berlow edits in https://github.com/googlefonts/googlesans/issues/284#issuecomment-877116548 with my minor edits to `Eng` and `eng` point definitions in the `GoogleSans-opsz18-wght380-GRAD0.ufo` master  to address an outline incompatibility issue that prevented compiles

### TODO

- [x] Rebase after #374 is merged (glyphsetdef conflict)
- [x] Remove Engtail (https://github.com/googlefonts/googlesans/pull/379/commits/18f7dc3196e1b20d346d27c773088f80fef289bf)
- [x] run `gs-normalize-designspace.py` on the sources, determine what still needs to be scrubbed from source files
- [ ] document glyph removal approach (#369)

PR #307, opened 2021-06-01
Glyph maintenance

madig:
Closes #122
Closes https://github.com/googlefonts/googlesans/issues/308
Closes https://github.com/googlefonts/googlesans/issues/152
Closes https://github.com/googlefonts/googlesans/issues/264
Closes https://github.com/googlefonts/googlesans/issues/280.

Also:

- cleans up [18] bracket layer directories

PR #565, opened 2023-01-06
Generate Google Sans Android build v10.001

chrissimpkins:
Google Sans v10.001 variable font update in Android U caused a significant font-related heap allocation size increase identified during  testing.

This appears to be due to the large file size increase that has occurred since the last GS release on Android.  We're working to diagnose the issue and determine if there is a way to address large fonts more optimally in the Android text rendering environment.  There won't be time for an Android-side fix before the deadline.

The last Google Sans Android release was the [v4.000 Roman and Italic variable fonts](https://github.com/googlefonts/googlesans/releases/tag/v4.000) that we distribute to all other font distributions. Those files will serve as our baseline for the change set in this PR.

We reached the following agreement with the Android team in order to land the update in Android U:

- subset Google Sans v10.001 back to the v4.000 glyph repertoire (this will allow us to support all bug fixes between v4 and v10.001 in glyphs that were present as of v4)
- add the glyph expansion requested by Android i18n as added in https://github.com/googlefonts/googlesans/pull/517 during Q4 2022

Furthermore, there is an Android-specific metric issue that led to visual regressions in their screenshot diff testing. We found that the regression is eliminated with the following:

- maintain head.yMin and head.yMax values as of the GS v4.000 release (i.e. mock in a way that they do not represent the actual compiler-calculated yMin and yMax)

I addressed two fontbakery fails in files that meet the above spec by:

- extension of the new glyph set to include `25CC`,  # DOTTED CIRCLE
- changing the OS/2.usWinAscent and OS/2.usWinDescent values to actual subset build yMax and yMin, respectively

Unfortunately, this will mean that we must now maintain a separate Android build target with a different glyph repertoire and likely expand language support in the Android environment with a different approach c/w our standard release process. We don't need to develop the build automation around this right now.  I am working with the Android engineering team to understand how we will need to approach language support expansion going forward.  We'll discuss Android V and beyond updates later in the year when we sort out how to approach it.

**For now, let's make sure that we are actually prod ready with the Android V targeted changes included in this PR.**

@MariannaPaszkowska Will you please review the new Android builds **in the attached archive below**? I haven't pushed the output files in this PR.  We can, but they are probably more appropriately distributed in GH releases.  They were built with the process defined by the Python script + requirements.txt file in this PR.  Please let me know if you identify any issues with the glyph repertoire or shaping in these files.  The glyph repertoire is based on the cmap table definition in the v4.000 builds + the OT feature support that was present as of v4.000.  I ran the files through hyperglot and they appear to include the expected target language support, including full coverage of the new languages added in  https://github.com/googlefonts/googlesans/pull/517.  I have not tested any layout, including the new Dutch shaping that was added in https://github.com/googlefonts/googlesans/pull/517.

@belluzj Can your team review the sources/fonts here to confirm that I have not missed something important in the subsetting and other binary edits that will be included in the release?

Thanks!


android/android-build.py
```diff
@@ -0,0 +1,195 @@
...
+        f"{v10_roman_input_filepath}",
+        f"--unicodes-file={roman_glyph_list_path}",
+        "--no-ignore-missing-glyphs",
+        "--notdef-outline",
+        "--layout-features+=c2sc,calt,case,ccmp,dlig,dnom,frac,jalt,liga,lnum,locl,numr,ordn,pnum,sinf,smcp,ss01,ss02,ss03,ss04,ss05,ss06,ss07,ss08,ss09,subs,sups,tnum,kern,mark,mkmk",
+        "--drop-tables= ",
+        "--name-IDs=*",
+        "--name-languages=*",
+        "--glyph-names",
+        "--recalc-bounds",
```
belluzj, line 88: @madig @RickyDaMa  go through the pyftsubset arguments and see if any could be missing (e.g. what about `--recalc-average-width`?)

android/android-build.py
```diff
@@ -0,0 +1,195 @@
...
+        "0198",  # LATIN CAPITAL LETTER K WITH HOOK
+        "0199",  # LATIN SMALL LETTER K WITH HOOK
+        "0186",  # LATIN CAPITAL LETTER OPEN O
+        "0254",  # LATIN SMALL LETTER OPEN O
+        "018F",  # LATIN CAPITAL LETTER SCHWA
+        "1E62",  # LATIN CAPITAL LETTER S WITH DOT BELOW
+        "1E63",  # LATIN SMALL LETTER S WITH DOT BELOW
+        "01B3",  # LATIN CAPITAL LETTER Y WITH HOOK
+        "01B4",  # LATIN SMALL LETTER Y WITH HOOK
+        "25CC",  # DOTTED CIRCLE (for combining diacritics) - fontbakery fail
```
belluzj, line 181: @madig @RickyDaMa double-check this list just in case

belluzj:
Hello @chrissimpkins, we'll have a look into this. I have two questions:

* Do you mean to commit the build artefacts (~TTFs and~ glyph lists)? Or should they be added to .gitignore?
  * I understand now that the TTFs are sources for the subsetting process, the actually subset fonts are indeed not pushed.
  * The glyph lists do come out of the build script so they could be ignored.
* ~Do you want us to call your new `android-build.py` script from the Makefile, and make sure the Android 10 TTFs are built and downloadable through GitHub actions?~
  * I just saw your sentence "We don't need to develop the build automation around this right now."

android/android-build.py
```diff
@@ -0,0 +1,195 @@
...
+        f"--output-file={roman_output_filepath}",
+    ]
+
+    # Italic subset options
+    italic_subset_options = [
+        f"{v10_italic_input_filepath}",
+        f"--unicodes-file={italic_glyph_list_path}",
+        "--no-ignore-missing-glyphs",
+        "--notdef-outline",
+        "--layout-features+=c2sc,calt,case,ccmp,dlig,dnom,frac,jalt,liga,lnum,locl,numr,ordn,pnum,sinf,smcp,ss01,ss02,ss04,ss05,ss06,ss07,ss08,ss09,subs,sups,tnum,kern,mark,mkmk",
```
madig, line 99: Is there a reason you specified features explicitly instead of saying `--layout-features=*`?

madig: A TTX dump comparison says you can go with `--layout-features=*` instead, no diff.

belluzj: Could it be "just in case" in the future Google Sans adds more features, but the Android team wants to keep only the glyphs associated with features that already existed in v4?

android/android-build.py
madig: Done, checks out.

android/android-build.py
```diff
@@ -0,0 +1,195 @@
...
+    # - yMin and yMax metrics in the subset fonts to the v4.000 values
+    # - Win Ascent and Win Descent metrics to actual y-min and -max values from compiler
+    # 2. Edit name table record ID5 to include "Android build"
+    for fontpath in [roman_output_filepath, italic_output_filepath]:
+        tt = TTFont(fontpath, recalcBBoxes=False)
+        head = tt["head"]
+        os2 = tt["OS/2"]
+
+        os2.usWinAscent = head.yMax
+        os2.usWinDescent = abs(head.yMin)
```
madig: The fontbakery code actually requires both values to be +1 yMin and yMax. Not sure if that is on purpose, but it actually makes the tests pass.

madig: Ah, this is actually our check code being wrong. The upstream check uses the correct "greater than or equal" approach.

madig:
The Mac-platform name table entries are dropped during subsetting. Most likely irrelevant for Android.

chrissimpkins:
Thanks Jany!

> * Do you mean to commit the build artefacts (~TTFs and~ glyph lists)? Or should they be added to .gitignore?

I'd say yes we can remove the binaries that were used as sources and implement from the release archives, but not a high priority right now. I think that the new build target should be added to our remote build process and that we should add new FB tests on the expectations for the Android builds.  I'll open issues up about this after we land this PR.

>  * The glyph lists do come out of the build script so they could be ignored.

Agree.  This can be part of the remote build when we set it up.  I pushed them to support an easier review of the Unicodes that we are using in the subsets.

> * ~Do you want us to call your new `android-build.py` script from the Makefile, and make sure the Android 10 TTFs are built and downloadable through GitHub actions?~

This is probably the proper direction.  Let's discuss it further after we wrap up the initial GS Flex release.

android/android-build.py
chrissimpkins: We only need the features that were defined in v4.  There have been a number of new features added to support shaping between v4 and v10 and it wasn't clear to me if you explicitly set them to keep with `*` that we maintain unnecessary shaping and glyph data that is outside of the target glyph repertoire.  It sounds like the subsetter does the "right" thing according to our expectations here as I understand Nikolaus' diff review comment above.

android/android-build.py
chrissimpkins: It can be either.  Probably makes sense to keep it consistent with our current prod version of GS which is +/- 1.  I'll change it. Ty!

madig:
I ran the regression testing suite over the fonts (excluding the Armenian, Bangla, Devanagari, Ethiopic, Georgian, Gujarati, Gurmukhi, Khmer, Lao, Odia, Sinhala, Tamil, Telugu and Thai ones) and the only failure is the missing Georgian Lari sign in the tnum tests. So, no known regressions :)

android/android-build.py
```diff
@@ -0,0 +1,195 @@
...
+    # Roman subset options
+    roman_subset_options = [
+        f"{v10_roman_input_filepath}",
+        f"--unicodes-file={roman_glyph_list_path}",
+        "--no-ignore-missing-glyphs",
+        "--notdef-outline",
+        "--layout-features+=c2sc,calt,case,ccmp,dlig,dnom,frac,jalt,liga,lnum,locl,numr,ordn,pnum,sinf,smcp,ss01,ss02,ss03,ss04,ss05,ss06,ss07,ss08,ss09,subs,sups,tnum,kern,mark,mkmk",
+        "--drop-tables= ",
+        "--name-IDs=*",
+        "--name-languages=*",
```
chrissimpkins, line 85: @madig Is there a way to address [the change in the name tables](https://github.com/googlefonts/googlesans/pull/565#issuecomment-1375656470) in the subsetter options?  Agree that we might not need to do so for future Android builds, but for now I think that it makes sense to keep the diff as small as possible assuming that this is simple.

chrissimpkins: I don't see anything that looks like an option to do this in https://github.com/fonttools/fonttools/blob/8c6e14674ace8e25ffbbc9162bd2fd065aafcc18/Lib/fontTools/subset/__init__.py#L327-L362

madig: I believe adding `--name-legacy` should do the trick.

android/android-build.py
chrissimpkins: Ty!

android/android-build.py
chrissimpkins: Ty!

2d525524ea1ae9ffc0cd1fb709a3411cec621735

chrissimpkins: Looks like it addressed this:

<details>
<summary>[Click for Diff]</summary>

```diff
--- v4/GoogleSans[GRAD,opsz,wght].ttf	2023-01-09T11:24:17.525189-05:00
+++ GoogleSans-Android.ttf	2023-01-09T11:27:26.728756-05:00
@@ -2,40 +2,40 @@
 <ttFont sfntVersion="\x00\x01\x00\x00" ttLibVersion="4.38">
 
   <name>
-    <namerecord nameID="272" platformID="1" platEncID="0" langID="0x0" unicode="True">
+    <namerecord nameID="274" platformID="1" platEncID="0" langID="0x0" unicode="True">
       Optical Size
     </namerecord>
-    <namerecord nameID="273" platformID="1" platEncID="0" langID="0x0" unicode="True">
+    <namerecord nameID="275" platformID="1" platEncID="0" langID="0x0" unicode="True">
       Max
     </namerecord>
-    <namerecord nameID="274" platformID="1" platEncID="0" langID="0x0" unicode="True">
+    <namerecord nameID="276" platformID="1" platEncID="0" langID="0x0" unicode="True">
       Text
     </namerecord>
-    <namerecord nameID="275" platformID="1" platEncID="0" langID="0x0" unicode="True">
+    <namerecord nameID="277" platformID="1" platEncID="0" langID="0x0" unicode="True">
       Weight
     </namerecord>
-    <namerecord nameID="276" platformID="1" platEncID="0" langID="0x0" unicode="True">
+    <namerecord nameID="278" platformID="1" platEncID="0" langID="0x0" unicode="True">
       Regular
     </namerecord>
-    <namerecord nameID="277" platformID="1" platEncID="0" langID="0x0" unicode="True">
+    <namerecord nameID="279" platformID="1" platEncID="0" langID="0x0" unicode="True">
       Medium
     </namerecord>
-    <namerecord nameID="278" platformID="1" platEncID="0" langID="0x0" unicode="True">
+    <namerecord nameID="280" platformID="1" platEncID="0" langID="0x0" unicode="True">
       Bold
     </namerecord>
-    <namerecord nameID="279" platformID="1" platEncID="0" langID="0x0" unicode="True">
+    <namerecord nameID="281" platformID="1" platEncID="0" langID="0x0" unicode="True">
       Grade
     </namerecord>
-    <namerecord nameID="280" platformID="1" platEncID="0" langID="0x0" unicode="True">
+    <namerecord nameID="282" platformID="1" platEncID="0" langID="0x0" unicode="True">
       Grade-50
     </namerecord>
-    <namerecord nameID="281" platformID="1" platEncID="0" langID="0x0" unicode="True">
+    <namerecord nameID="283" platformID="1" platEncID="0" langID="0x0" unicode="True">
       0
     </namerecord>
-    <namerecord nameID="282" platformID="1" platEncID="0" langID="0x0" unicode="True">
+    <namerecord nameID="284" platformID="1" platEncID="0" langID="0x0" unicode="True">
       Grade200
     </namerecord>
-    <namerecord nameID="283" platformID="1" platEncID="0" langID="0x0" unicode="True">
+    <namerecord nameID="285" platformID="1" platEncID="0" langID="0x0" unicode="True">
       Italic
     </namerecord>
     <namerecord nameID="0" platformID="3" platEncID="1" langID="0x409">
@@ -48,13 +48,13 @@
       Regular
     </namerecord>
     <namerecord nameID="3" platformID="3" platEncID="1" langID="0x409">
-      4.000;GOOG;GoogleSans-Regular
+      10.001;GOOG;GoogleSans-Regular
     </namerecord>
     <namerecord nameID="4" platformID="3" platEncID="1" langID="0x409">
       Google Sans Regular
     </namerecord>
     <namerecord nameID="5" platformID="3" platEncID="1" langID="0x409">
-      Version 4.000
+      Version 10.001;[06685905];Android build
     </namerecord>
     <namerecord nameID="6" platformID="3" platEncID="1" langID="0x409">
       GoogleSans-Regular
@@ -75,13 +75,13 @@
       Google offers many fonts on open source terms. Google Sans is not one of them. Please see google.com/fonts for alternatives.
     </namerecord>
     <namerecord nameID="256" platformID="3" platEncID="1" langID="0x409">
-      File Size Units
+      Number Pad Asterisk
     </namerecord>
     <namerecord nameID="257" platformID="3" platEncID="1" langID="0x409">
-      Number pad asterisk
+      Colon Time
     </namerecord>
     <namerecord nameID="258" platformID="3" platEncID="1" langID="0x409">
-      Colon time
+      File Size Units
     </namerecord>
     <namerecord nameID="259" platformID="3" platEncID="1" langID="0x409">
       Alternate Arrows
@@ -93,69 +93,75 @@
       iota adscript
     </namerecord>
     <namerecord nameID="262" platformID="3" platEncID="1" langID="0x409">
+      Single Story a
+    </namerecord>
+    <namerecord nameID="263" platformID="3" platEncID="1" langID="0x409">
+      Alternate Symbols
+    </namerecord>
+    <namerecord nameID="264" platformID="3" platEncID="1" langID="0x409">
       Thin Numbers
     </namerecord>
-    <namerecord nameID="263" platformID="3" platEncID="1" langID="0x409">
+    <namerecord nameID="265" platformID="3" platEncID="1" langID="0x409">
       Size
     </namerecord>
-    <namerecord nameID="264" platformID="3" platEncID="1" langID="0x409">
+    <namerecord nameID="266" platformID="3" platEncID="1" langID="0x409">
       Weight
     </namerecord>
-    <namerecord nameID="265" platformID="3" platEncID="1" langID="0x409">
+    <namerecord nameID="267" platformID="3" platEncID="1" langID="0x409">
       Grade
     </namerecord>
-    <namerecord nameID="266" platformID="3" platEncID="1" langID="0x409">
+    <namerecord nameID="268" platformID="3" platEncID="1" langID="0x409">
       Regular
     </namerecord>
-    <namerecord nameID="267" platformID="3" platEncID="1" langID="0x409">
+    <namerecord nameID="269" platformID="3" platEncID="1" langID="0x409">
       Medium
     </namerecord>
-    <namerecord nameID="268" platformID="3" platEncID="1" langID="0x409">
+    <namerecord nameID="270" platformID="3" platEncID="1" langID="0x409">
       Bold
     </namerecord>
-    <namerecord nameID="269" platformID="3" platEncID="1" langID="0x409">
+    <namerecord nameID="271" platformID="3" platEncID="1" langID="0x409">
       Text Regular
     </namerecord>
-    <namerecord nameID="270" platformID="3" platEncID="1" langID="0x409">
+    <namerecord nameID="272" platformID="3" platEncID="1" langID="0x409">
       Text Medium
     </namerecord>
-    <namerecord nameID="271" platformID="3" platEncID="1" langID="0x409">
+    <namerecord nameID="273" platformID="3" platEncID="1" langID="0x409">
       Text Bold
     </namerecord>
-    <namerecord nameID="272" platformID="3" platEncID="1" langID="0x409">
+    <namerecord nameID="274" platformID="3" platEncID="1" langID="0x409">
       Optical Size
     </namerecord>
-    <namerecord nameID="273" platformID="3" platEncID="1" langID="0x409">
+    <namerecord nameID="275" platformID="3" platEncID="1" langID="0x409">
       Max
     </namerecord>
-    <namerecord nameID="274" platformID="3" platEncID="1" langID="0x409">
+    <namerecord nameID="276" platformID="3" platEncID="1" langID="0x409">
       Text
     </namerecord>
-    <namerecord nameID="275" platformID="3" platEncID="1" langID="0x409">
+    <namerecord nameID="277" platformID="3" platEncID="1" langID="0x409">
       Weight
     </namerecord>
-    <namerecord nameID="276" platformID="3" platEncID="1" langID="0x409">
+    <namerecord nameID="278" platformID="3" platEncID="1" langID="0x409">
       Regular
     </namerecord>
-    <namerecord nameID="277" platformID="3" platEncID="1" langID="0x409">
+    <namerecord nameID="279" platformID="3" platEncID="1" langID="0x409">
       Medium
     </namerecord>
-    <namerecord nameID="278" platformID="3" platEncID="1" langID="0x409">
+    <namerecord nameID="280" platformID="3" platEncID="1" langID="0x409">
       Bold
     </namerecord>
-    <namerecord nameID="279" platformID="3" platEncID="1" langID="0x409">
+    <namerecord nameID="281" platformID="3" platEncID="1" langID="0x409">
       Grade
     </namerecord>
-    <namerecord nameID="280" platformID="3" platEncID="1" langID="0x409">
+    <namerecord nameID="282" platformID="3" platEncID="1" langID="0x409">
       Grade-50
     </namerecord>
-    <namerecord nameID="281" platformID="3" platEncID="1" langID="0x409">
+    <namerecord nameID="283" platformID="3" platEncID="1" langID="0x409">
       0
     </namerecord>
-    <namerecord nameID="282" platformID="3" platEncID="1" langID="0x409">
+    <namerecord nameID="284" platformID="3" platEncID="1" langID="0x409">
       Grade200
     </namerecord>
-    <namerecord nameID="283" platformID="3" platEncID="1" langID="0x409">
+    <namerecord nameID="285" platformID="3" platEncID="1" langID="0x409">
       Italic
     </namerecord>
   </name>
```

</details>

Ty!

chrissimpkins:
> I ran the regression testing suite over the fonts (excluding the Armenian, Bangla, Devanagari, Ethiopic, Georgian, Gujarati, Gurmukhi, Khmer, Lao, Odia, Sinhala, Tamil, Telugu and Thai ones) and the only failure is the missing Georgian Lari sign in the tnum tests. So, no known regressions :)

Great news!

chrissimpkins:
> > I ran the regression testing suite over the fonts (excluding the Armenian, Bangla, Devanagari, Ethiopic, Georgian, Gujarati, Gurmukhi, Khmer, Lao, Odia, Sinhala, Tamil, Telugu and Thai ones) and the only failure is the missing Georgian Lari sign in the tnum tests. So, no known regressions :)
> 
> Great news!

Do we have regression tests built in around the new Dutch shaping?

android/android-build.py
chrissimpkins: add323db9529ff82cc8d6676b03716ef4db64aff

confirmed in new builds.  Ty!

chrissimpkins:
Added a zip archive with new VF builds as of commit https://github.com/googlefonts/googlesans/pull/565/commits/add323db9529ff82cc8d6676b03716ef4db64aff in the first post.  

Also linked below:

https://github.com/googlefonts/googlesans/files/10375181/android-v10.001-add323d.zip

madig:
I believe you are looking for the tests added in https://github.com/googlefonts/googlesans/pull/517/files? Yes, they were run.

MariannaPaszkowska:
I can confirm that glyphs from the https://github.com/googlefonts/googlesans/pull/517 are present.
I noticed that some glyphs from v4 are missing in this build, and some glyphs might not be necessary for this set (localized for Tamil). I've included more details below. 


glyphs from 517:
```
ĳ LATIN CAPITAL LIGATURE IJ (U+0132) 
Ĳ LATIN SMALL LIGATURE IJ (U+0133) 
Ƙ LATIN CAPITAL LETTER K WITH HOOK (U+0198)
Ƴ LATIN CAPITAL LETTER Y WITH HOOK (U+01B3)
Ɓ LATIN CAPITAL LETTER B WITH HOOK (U+0181)
Ɗ LATIN CAPITAL LETTER D WITH HOOK (U+018A)
ƙ LATIN SMALL LETTER K WITH HOOK (U+0199)
ƴ LATIN SMALL LETTER Y WITH HOOK (U+01B4)
ɓ LATIN SMALL LETTER B WITH HOOK (U+0253)
ɗ LATIN SMALL LETTER D WITH HOOK (U+0257)
Ṣ LATIN CAPITAL LETTER S WITH DOT BELOW (U+1E62)
ṣ`LATIN SMALL LETTER S WITH DOT BELOW (U+1E63)
Ɔ LATIN CAPITAL LETTER OPEN O (U+0186)
Ɛ LATIN CAPITAL LETTER OPEN E (U+0190)
ɔ LATIN SMALL LETTER OPEN O (U+0254)
ɛ LATIN SMALL LETTER OPEN E (U+025B)
Ⓖ CIRCLED LATIN CAPITAL LETTER G (U+24BC)
```

also additional compared to v4:
cap characters (might be good to keep them):
```
guillemotleft.ca... -70538              «                   
guillemotright.c... -70880              »                   
guilsinglleft.ca... -35269              ‹                   
guilsinglright.c... -35440              ›  
```

loclTAML glyphs - those might not be needed: 

```
asterisk.loclTAM... -53746              *                   
braceleft.loclTA... -92316              {                   
braceright.loclT... -92252              }                   
bracketleft.locl... -98464              [                   
bracketright.loc... -98464              ]                   
colon.loclTAML      -25838              :                   
comma.loclTAML      -18381              ,                   
eight.loclTAML      -132389             8                   
eight.loclTAML.t... -126387             8                   
ellipsis.loclTAM... -38574              …                   
exclam.loclTAML     -54154              !                   
five.loclTAML       -114487             5                   
five.loclTAML.tf... -110004             5                   
four.loclTAML       -107102             4                   
four.loclTAML.tf... -103369             4   
nine.loclTAML       -101253             9                   
nine.loclTAML.tf... -99343              9                   
one.loclTAML        -60999              1                   
one.loclTAML.tf     -63195              1                   
parenleft.loclTA... -79921              (                   
parenright.loclT... -79921              )                   
period.loclTAML     -12858              .                   
question.loclTAM... -83977              ?                   
quotedblleft.loc... -36763              “                   
quotedblright.lo... -36763              ”                   
quoteleft.loclTA... -18381              ‘                   
quoteright.loclT... -18381              ’                   
semicolon.loclTA... -31361              ;                   
seven.loclTAML      -75941              7                   
seven.loclTAML.t... -73499              7                   
six.loclTAML        -101110             6                   
six.loclTAML.tf     -99171              6                   
slash.loclTAML      -55848              /                   
three.loclTAML      -106396             3                   
three.loclTAML.t... -101926             3                   
two.loclTAML        -105718             2                   
two.loclTAML.tf     -101541             2  
```

glyphs from v4 that seem to be missing in this build:

```
r_f
r_f_f
r_t
i.loclTRK.sc
tonos.case
uni1FBF.case
uni1FBD.case
uni1FFE.case
uni1FDD.case
uni1FCE.case
uni1FDE.case
uni1FCF.case
uni1FDF.case
uni1FED.case
uni1FEE.case
uni1FC1.case
uni1FEF.case
uni1FFD.case
uni1FC0.case
uni1FCD.case
Glogo
ologo
glogo
llogo
elogo
Googlelogo
finalkafdagesh_shevahb
finalkafdagesh_qamatshb
finalkaf_shevahb
finalkaf_qamatshb
lamed_holamhb
lamed_dagesh_holamhb
hatafsegol_siluqlefthb
hatafpatah_siluqlefthb
hatafqamats_siluqlefthb
Engtail
Germandbls
Germandbls.sc
rlig
uni05E3.BRACKET.18
uniFB43.BRACKET.18
uni05E4.BRACKET.18
uniFB44.BRACKET.18
uniFB4E.BRACKET.18
uni05E1.BRACKET.18
uniFB41.BRACKET.18
```

madig:
The TAML glyphs might be there because they're reachable via the normal punctuation. If you really want them gone, try `--layout-scripts-=tml2.dflt` etc.

chrissimpkins:
Thanks Marianna!

locl.TAML - I suspect this is because we do want locl and they are based on encoded glyphs that we keep in the subset.

r-ligatures - these were removed last year.  OK that they are gone.

logo forms - it looks like they were all renamed (e.g. Glogo to G.logo)?  the logo ligatures work [*edit: confirmed*]

case forms - need to look into what characters these are and why they are missing.  We must have removed them at some stage?   [*edit: these are Greek glyphs, confirmed that they were intentionally removed. Irene validated shaping post-removal*]

Hebrew chars - don't recall these changes, will need to investigate [*edit: glyphs were renamed, confirmed*]

Engtail / Germandbls - don't know why [*edit: Engtail intentionally removed, Germandbls was renamed, confirmed*]

*.BRACKET.18 - all ok that they are gone. These were the old rvrn forms

android/android-build.py
chrissimpkins: Thoughts here?

chrissimpkins:
Also agree with keeping the guillemot / guilsing case forms

chrissimpkins:
All of the Greek (tonos.case + uni*.case) glyphs were intentionally removed in https://github.com/googlefonts/googlesans/pull/391.  Irene confirmed that this is ok and that we are shaping as expected in Greek without them

chrissimpkins:
The Hebrew ligatures (?), i.loclTRK.sc, and Germandbls (including smcp) all had glyph name changes documented in https://github.com/googlefonts/googlesans/commit/5f23b521c753d490f0b0721f23acc507ef0e5dde

They should still be present but with different glyph names

chrissimpkins:
Engtail was intentionally removed in https://github.com/googlefonts/googlesans/pull/379

chrissimpkins:
Duplicate Google logo forms were intentionally removed in https://github.com/googlefonts/googlesans/pull/307

chrissimpkins:
I think that we have a reasonable explanation for all of the unexpected v4 era glyph removals that Marianna noted above.

chrissimpkins:
> The TAML glyphs might be there because they're reachable via the normal punctuation. If you really want them gone, try `--layout-scripts-=tml2.dflt` etc.

This didn't seem to modify the glyph contents

android/android-build.py
chrissimpkins: Added `--recalc-average-width` in 0a290350a080f30a3a44b9b8f9cbd8efddae960b

It addresses a fontbakery warning.  Fontbakery seems to think that it is the right thing to do. :)

madig: Hm, my comment got lost somewhere... maybe also add `--recalc-max-context`.

madig:
Try `--layout-scripts=latn,grek,cyrl,hebr`. Those seem to be the only scripts supported by v4.

android/android-build.py
chrissimpkins: Thanks Nikolaus.  This option does not appear to change anything in the binaries.

madig: Ok. No other option stood out to me when skimming them.

chrissimpkins:
> Try `--layout-scripts=latn,grek,cyrl,hebr`. Those seem to be the only scripts supported by v4.

This appears to address it.  There is a large GPOS and GSUB table diff.  I ran all of the general/Latin/Cyrillic/Greek/Hebrew shaping tests again and, with the exception of the Georgian lira symbol that is missing in the tnum tests, all looks well.  A quick glance through the glyf diff suggests that all of the loclTAML glyphs are removed.  There are a large number of other localized glyphs for other scripts that go away too.  I'll push the update after I confirm that all of the removed glyphs are as expected.

Thank you very much Nikolaus.  This information was extremely helpful.

chrissimpkins:
glyf table diff LGTM

Diff pre/post the layout-scripts addition is attached:

[glyf_diff.diff.zip](https://github.com/googlefonts/googlesans/files/10395045/glyf_diff.diff.zip)

chrissimpkins:
converted to post v3 table for these Android-only builds in 2b6199e3dbea20faa1c6c2a90dd35bde9a5c55da

It looks like it decreases file size by ~15 - 20 kb per VF build

chrissimpkins:
File sizes as of https://github.com/googlefonts/googlesans/pull/565/commits/2b6199e3dbea20faa1c6c2a90dd35bde9a5c55da

```
GoogleSans-Android.ttf

             Summary Statistics
┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━┓
┃                    Statistic ┃  Size (B) ┃
┡━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━┩
│                        Total │ 588260.00 │
│ Avg per encoded glyph (1330) │    442.30 │
│   Avg per total glyph (2424) │    242.68 │
└──────────────────────────────┴───────────┘
```

```
GoogleSans-Android-Italic.ttf

             Summary Statistics
┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━┓
┃                    Statistic ┃  Size (B) ┃
┡━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━┩
│                        Total │ 611668.00 │
│ Avg per encoded glyph (1330) │    459.90 │
│   Avg per total glyph (2420) │    252.76 │
└──────────────────────────────┴───────────┘

```

chrissimpkins:
Latest build files @ https://github.com/googlefonts/googlesans/pull/565/commits/c4ebc1cdd8a6d085ce4fecb4996fa4dbd4a7b52c

[GoogleSans-Android-build.zip](https://github.com/googlefonts/googlesans/files/10396146/GoogleSans-Android-build.zip)

chrissimpkins:
I spoke with the Pixel engineering team and they push to their system image with the file paths:

- GoogleSans-Regular.ttf
- GoogleSans-Italic.ttf

c4ebc1cdd8a6d085ce4fecb4996fa4dbd4a7b52c updates the file paths so that we build to their shipping file path spec.  Updated the font zip archive in the above post with the builds at this commit.

MariannaPaszkowska (approved):

chrissimpkins:
Marianna approved the builds.

@madig Are we OK to ship from your standpoint?

madig:
I believe so.

chrissimpkins:
Thanks all!

I shipped the files compiled at https://github.com/googlefonts/googlesans/pull/565/commits/c4ebc1cdd8a6d085ce4fecb4996fa4dbd4a7b52c to the Pixel eng team.  

I'll add README documentation with background / changes included in this build target before we merge this.  Will also prep a GH release that includes the Android build target files.

chrissimpkins:
Will keep open until we address https://github.com/googlefonts/googlesans/issues/566

MariannaPaszkowska:
@chrissimpkins Here are the static instances of GS Android build v10.001 [GoogleSans-Android-singleFonts.zip](https://github.com/googlefonts/googlesans/files/11100518/GoogleSans-Android-singleFonts.zip)

Here is what testing with the latest FontBakery showed:
1. The opentype profile v0.8.11: 
WARN: some glyphs have coordinates that are out of bounds _(in 4 fonts out of 12)_ 
WARN: the post table format should be updated from 2.5 to 3 or downgraded to 2. _(we should add this in the future releases of GS)_

2. Font bakery report - googlefonts profile v0.8.11
Design:
• FAIL missing code points 0x01CE SMALL LETTER A WITH CARON - _(we should add this glyph to the future releases of GS)_

Shaping: 
• FAIL The dot of soft dotted characters used in orthographies must disappear in the following strings: į̀ į́ į̂ į̃ į̄ į̌ і́ and should disappear in other cases, for example: i̒ i̛̒ ị̒ i̤̒ i̦̇ i̦̊ i̦̋ ǐ̦ i̦̒ i̧̒ i̮̒ i̱̒ j̒ j̛̒ j̣̒ j̤̒ j̦̀ j̦́ j̦̃ j̦̄  _(we should improve that for the future releases of GS)_
• FAIL The following unacceptable control characters were identified: uni0002, uni0009 _(we should consider removing those for future releases)_

madig:
Comparing the Chris's VFs against the newly cut statics, I found:

* The statics have a PANOSE value set. Good.
* Overlapping composites are decomposed, driving up file size
* Inscrutable GPOS diffs
* Some glyphs have a different left side bearing value in the hmtx table. Probably irrelevant, because it doesn't impact shaping.
* Vertical metrics numbers are the same and I can't immediately spot diffs there between Chris' and Marianna's fonts in a browser test page

Compared to v4.000, I found:

* The head table font bounding box can differ between v4 and the newly cut instances. This may lead to vertical metrics diffs on Android, unless the developers manually control line spacing in apps.
* Various glyphs have different spacing in v4 vs. the newly cut instances, changing shaping.
* The newly cut instances don't use production names (post table 3.0). This is probably only interesting for PDF generation and having autohinting on macOS. It makes the diffs less scrutable though.
* v4 had some different OS/2 win metrics, which may impact line spacing in some enviroments.

Nothing interetsing found comparing to v10.001, but the large diffs make it very hard to spot interesting diffs.

Fontbakery (googlefonts profile) complains that:

1. GoogleSansText-Medium(Italic) have too long a name, though this seems to be a Windows issue
2. There's no meta table (never seen that one)
3. OS/2.usWinAscent and -Descent are too small (maybe irrelevant for Android?)
4. There are a bunch of unreachable glyphs
5. The font shave a soft hyphen
6. Some math signs don't have the same width (design issue)
7. Various jaggy segments and semi-vertical or semi-horizontal lines (design issue)
8. Some glyphs have a faulty bounding box
9. The stuff Marianna found

The unreachable glyphs and glyph bounding box issues can be made to go away by running the pruner over them (merge https://github.com/googlefonts/googlesans/pull/570 first, then run `python scripts/internal/prune_font_binary.py path/to/fonts/*.ttf`).

chrissimpkins:
@madig I'd like to look into how we might integrate these Android-targeted VF and static builds into the release archives going forward.  Obviously not necessary for the previous v10.001 release that this PR is based on, but we'll need to do this again and I think that we should consider this a new build target in our releases.  I'm looking into how Android imagines releasing scripts going forward.  Let's allow those discussions to play out before we decide how to address this.

chrissimpkins:
Marianna reviewed the new statics and Nikolaus' review of the new statics.  She gave the green light to ship these files to Eng for onboarding into the GMS core distribution.

madig:
I added a `build-android` make target, so adding them to a release Zip is straightforward.

chrissimpkins:
Discussing the need to maintain these build targets with the Android SysUI team.

chrissimpkins:
We're attempting to onboard the full, default GS family as of Android V.  Ideally, we will no longer need to maintain this subset build target. I'll update here if this lands.

chrissimpkins:
We aim to attempt a new Android release with our default v12.000 release files this year.  Keeping this open until we confirm that it lands.

belluzj:
Hello @chrissimpkins , anything for us to do relating to your last message?

chrissimpkins:
Nothing to do.  Just an update on the status of this branch.  Let's continue to keep it for now.  But hopefully it will go away this year.  Thanks Jany.

MariannaPaszkowska:
Updated PR here #711

Issue #542, opened 2022-10-09
Add Gurage support


chrissimpkins:

chrissimpkins:
@NeilSureshPatel Will you please provide a tentative source delivery date when you have a chance? Ty!

NeilSureshPatel:
I'll touch base with Mark tomorrow and get you a tentative date

NeilSureshPatel:
@chrissimpkins It looks like we could deliver as soon as tomorrow or early next week.

NeilSureshPatel:
@chrissimpkins Quick question about delivering sources. In the past we had put the sources in my [google-sans-ethiopic](https://github.com/NeilSureshPatel/google-sans-ethiopic/tree/master/source/GoogleSans) repo. A lot has changed since then. Would it be better to fork this repo and upload the Gurage updates? Also, should we convert them to UFO or will that happen during the onboarding. I assume we will need to update import_glyphs.txt as well. Thanks.

chrissimpkins:
It is ok to drop the Glyphs format sources with shaping code, glyph name import list, and text runs required for the regression tests in a zip archive in this issue thread.  Make it simple on yourself. It would be worth **storing** them in one of your git repos for safe keeping and future reference.  But we don't need a repo to transfer the data and import.  Thanks!

MarkJamra:
Google Sans Ethiopic Gurage is attached.
[GoogleSans_Ethiopic_Gurage.zip](https://github.com/googlefonts/googlesans/files/9840922/GoogleSans_Ethiopic_Gurage.zip)

chrissimpkins:
Thank you very much Mark!

@belluzj @MariannaPaszkowska Let's discuss timing on this relative to other scripts on Monday.

madig:
During conversion to UFOs, I'm getting a message about how the public.kern1.ethi_hhwagurage kerning group does not exist. Is this a leftover?

madig:
Also, the fea file does not compile because blocks are missing their name at the end, like in `feature bla { ... } bla;`, what editor are you using?

madig:
Also, the import glyph list is incomplete. I had to add geez part 11 or some such, and it now complains about ethi_num (referenced in ethiopic.fea). Can you please double check the list? Or is ethi_num the feature group name? If so, can you please fix the feature file to compile when using fontmake?

NeilSureshPatel:
[...]

PR #574, opened 2023-04-11
Import Gurage

madig:
Related: https://github.com/googlefonts/googlesans/issues/542


MariannaPaszkowska:
Will be released here: https://github.com/googlefonts/googlesans-gurage

PR #578, opened 2023-04-28
Fix soft-dotted characters

### LINKED ISSUE #576: Fix the soft dotted characters (make the dot dissapear)
Fix the following strings: į̀ į́ į̂ į̃ į̄ į̌ і́

see fail: https://github.com/googlefonts/googlesans/actions/runs/4796497188/jobs/8533124125#step:5:524

belluzj:
@MariannaPaszkowska Fixes #576 

Before:
image

After:
image

I have only tested on a single UFO locally, I'll wait for the CI build of the variable font to check that it works in all the designspace (no reason not to since the GSUB feature is global).

I also added the je-cy into the mix, as it's listed on this page: https://googlefonts.github.io/gf-guide/diacritics.html#soft-dotted-glyphs


belluzj:
I added a second commit in which I copied the ogonek anchor from the /idotless to /i, so that the combo /i/ogonekcomb that's introduced by feature code looks OK:

Before the second commit:
image

After:
image

belluzj:
I had a look at the upright and italic VFs with the following test string: į į̀ į́ į̂ į̃ į̄ į̌ і́ ј́  and all corners of the designspace look fine to me. I'll fix the build failures.

moyogo:
You don't have to decompose idotbelow or iogonek all the time, they can be decomposed when combined with a top mark either in its own lookup block as now or even moved in the soft dotted lookup block. Decomposing them all the time means their kerning is always ignored.

What about interaction with other features like smallcaps? 

For future proof soft dotted lookup, you can use `lookupflag UseMarkFilteringSet @CombiningTopAccents;` instead of having a sub with combining non top accents.

belluzj:
The small caps work fine at the moment. I didn't consider the kerning problem, thanks for pointing that out. @MariannaPaszkowska should I change the feature code to use Denis's suggestion of the UseMarkFilteringSet ? Let's discuss tomorrow

moyogo:
@belluzj there seems to be a problem with "fį" and the interaction with the liga feature.

EDIT: there was already an issue with "fị" in italic, but now this affects both idotbelow and iogonek.
Screenshot 2023-05-02 at 11 27 34

belluzj:
Marianna: we should try to use UseMarkFilteringSet if possible like Denis suggests.

belluzj:
I pushed a new version that uses UseMarkFilteringSet. Here are screenshots of the test string: į į̀ į́ į̂ į̃ į̄ į̌ і́ ј́ fį fị

Upright<br>
image

Upright small caps<br>
image

Italic<br>
image

Italic small caps<br>
image

PR #579, opened 2023-06-05
Allow inconsistent family names

madig:


madig:
@MariannaPaszkowska we need this to make all future builds pass.

qa/check-googlesans.py
```diff
@@ -62,6 +62,8 @@
     "com.google.fonts/check/varfont/bold_wght_coord",  # Buggy in 0.8.9
     # https://github.com/googlefonts/googlesans/issues/576:
     "com.google.fonts/check/soft_dotted",
+    # We intentionally set the family names as they are
```
chrissimpkins, line 65: There have been no changes to our previous production approach to naming?  What is the issue that FB is flagging here and should we be addressing it at the FB level rather than ignoring here?

madig: The family names are wrong as per your decision somewhere (historical accident etc). This is a new FB check that flags it.

madig: Here's the check output:

```
Rationale:                                                                
                                                                            
Per the OpenType spec: * "...many existing applications that use this     
pair of names assume that a Font Family name is shared by at most four    
fonts that form a font style-linking group" * "For extended typographic   
families that includes fonts other than the four basic styles(regular,    
italic, bold, bold italic), it is strongly recommended that name IDs 16   
and 17 be used in fonts to create an extended, typographic grouping." *   
"If name ID 16 is absent, then name ID 1 is considered to be the          
typographic family name."                                                 
                                                                            
https://learn.microsoft.com/en-us/typography/opentype/spec/name           
                                                                            
Fonts within a font family all must have consistent names in the          
Typographic Family name (nameID 16) or Font Family name (nameID 1),       
depending on which it uses.                                               
                                                                            
Inconsistent font/typographic family names across fonts in a family can   
result in unexpected behaviors, such as broken style linking.             
                                                                            

FAIL 2 different Font Family names were found:                             
                                                                            
    • 'Google Sans' was found in:                                        
        • GoogleSans-Medium.ttf (nameID 16)                               
        • GoogleSans-Italic.ttf (nameID 1)                                
        • GoogleSans-Bold.ttf (nameID 1)                                  
        • GoogleSans-MediumItalic.ttf (nameID 16)                         
        • GoogleSans-BoldItalic.ttf (nameID 1)                            
        • GoogleSans-Regular.ttf (nameID 1)                               
    • 'Google Sans Text' was found in:                                   
        • GoogleSansText-Regular.ttf (nameID 1)                           
        • GoogleSansText-Bold.ttf (nameID 1)                              
        • GoogleSansText-BoldItalic.ttf (nameID 1)                        
        • GoogleSansText-Italic.ttf (nameID 1)                            
        • GoogleSansText-Medium.ttf (nameID 16)                           
        • GoogleSansText-MediumItalic.ttf (nameID 16) [code: inconsistent-family-name]   
```

chrissimpkins: That is intentional.  The statics should have different family names.  Thanks Nikolaus.

madig:
Grm, this doesn't silence them on statics :/

madig:
@felipesanches may I ping you here? Can you tell me why fontbakery does not use our exclude list?

* List: https://github.com/googlefonts/googlesans/blob/main/qa/check-googlesans.py#L50-L67
* Filter: https://github.com/googlefonts/googlesans/blob/main/qa/check-googlesans.py#L533-L541
* Command: `fontbakery check-profile --auto-jobs -C --loglevel WARN qa/check-googlesans.py build/GoogleSans/static/*.ttf`
    * You can get static font binaries from https://github.com/googlefonts/googlesans/suites/13870143441/artifacts/770895536

madig:
It seems the check filter isn't even called.

madig:
Ah. It is, but `font` is None, so the `if` doesn't actually trigger properly.

felipesanches:
Sorry, I only saw your messages now. Do you still need help here?

madig:
Nope, solved it, thanks :)

PR #584, opened 2023-06-27
Use upstream check for GRAD not reflowing layout

### LINKED ISSUE #253: Add QA check for differences in advance widths and kerning across the GRAD axis masters

madig:
This promptly found 

* Men_Ech-arm.liga
* baht.tf
* semicolon.loclARM and    
* men_xeh-arm.liga.sc

 in violation of the advance width rules :)

Closes https://github.com/googlefonts/googlesans/issues/253


Hoolean:
The ERROR'ing tests on this PR will be fixed as fonttools/fontbakery#4357 is released; aside from this, to close this issue we should:

- [x] Investigate difference between the glyphs reported by the test locally and on the CI
  - Potentially an environment issue with my machine, but we should confirm this
- [x] Open an issue with the full list of glyphs reported
  - Both to verify against Nikolaus' list above, and because the full glyphs list from the CI is truncated
  - Fix with Chris's script `gs-fix-grad-widths.py`
  - Ask Eduardo to double-check the list of glyphs to make they're optically centred in the GRAD

Hoolean:
@MariannaPaszkowska Ready for review and merge 🎉

This PR has been rebased and the check now gives the same results on the CI and locally, including as a subset the glyphs that Nikolaus flagged above.

(the two failing tests are intended, as they are from issues identified by this new check; we have opened a PR at #609 to begin fixing these)

PR #603, opened 2023-10-19
More flexible gs-stat.py

simoncozens:
It makes sense for "standalone" GS versions (Loopless, Arabic, etc.) to use the same tooling as main GS as much as possible. In particular, it would be good to sync the generation of STAT tables so we all get the expected values. However, some of the standalone versions don't have all the axes that main GS has, or don't have italic counterparts. This patch allows them to use `gs-stat.py` to generate their STAT tables, but to omit the `GRAD` and `opsz` portions of STAT if those axes are not in the font; also it allows them to skip the `ital` axis if `--no-italic` is passed.


belluzj (approved):
LGTM

PR #609, opened 2023-12-15
Stop GRAD reflow with gs-fix-grad-widths.py

Hoolean:
This PR automatically adjusts the width of some glyphs in the GRAD masters (see #584) to prevent reflow as the GRAD axis is varied. It is marked as WIP as the changes will require manual review to ensure that the glyphs are still centred and spaced correctly.

In addition, the 'fix GRAD widths' script has been lightly updated to work with newer versions of ufoLib2.


Hoolean:
@MariannaPaszkowska Ready for review: if we merge this into main then Eduardo can review the affected glyphs as part of his work on re-spacing 🚀

MariannaPaszkowska:
@belluzj Eduardo can start his review, many thanks!

PR #615, opened 2024-01-09
Tweak avar to fix opsz 18 on Firefox MacOS

### LINKED ISSUE #462: `opsz` axis is not working on Firefox + macOS (only)
I‘m not sure if this is a known issue, but Google Sans `opsz` axis does not seem to be working reliably on Firefox (for macOS). I noticed this first with Google Sans Arabic, but found the same issue with Google Sans, so I wonder if it is something related to the design space setup or the build pipeline.

Firefox (above `opsz `is set to 18, below 17):
Firefox

Chrome:
Chrome

RickyDaMa:
Fixes #462 

I had a look at the [Firefox bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1758147) and also figured I could try twiddling the numbers a bit and seeing if it would work. Lo and behold, we get Google G now in Firefox on MacOS at opsz 18!

### DIFF
--- a/source/GoogleSans/GoogleSans-Italic.designspace
+++ b/source/GoogleSans/GoogleSans-Italic.designspace
@@ -3,7 +3,7 @@
   <axes>
     <axis tag="opsz" name="Size" minimum="17" maximum="18" default="18">
       <map input="17" output="17"/>
-      <map input="17.99" output="17.1"/>
+      <map input="17.9" output="17.11"/>
       <map input="18" output="18"/>
     </axis>
     <axis tag="wght" name="Weight" minimum="400" maximum="700" default="400">
--- a/source/GoogleSans/GoogleSans.designspace
+++ b/source/GoogleSans/GoogleSans.designspace
@@ -3,7 +3,7 @@
   <axes>
     <axis tag="opsz" name="Size" minimum="17" maximum="18" default="18">
       <map input="17" output="17"/>
-      <map input="17.99" output="17.1"/>
+      <map input="17.9" output="17.11"/>
       <map input="18" output="18"/>
     </axis>
     <axis tag="wght" name="Weight" minimum="400" maximum="700" default="400">


chrissimpkins:
Fantastic! No impact in other environments?

RickyDaMa:
I can confirm it still works as expected on Chrome on MacOS, but I haven't tested anything else yet

chrissimpkins:
@m4rc1e would diffbrowsers work here to confirm that we aren't creating new issues across environments to fix the Firefox issue? Can we run a test across environments using different values on the optical size axis?

m4rc1e:
Currently, Diffenator2's diffbrowser can only change strings atm. This does sound like an incredibly useful feature imo. I can add it next week? it shouldn't be too difficult.

m4rc1e:
I've just added a pr to diffenator2 which allows you to provide your own html templates, https://github.com/googlefonts/diffenator2/pull/125. Once merged, I'll make a html template which tests this opsz feature.

m4rc1e:
I'm able to replicate Khaled's findings with the latest release. I'll build this pr and see what we get. Just realised diffenator2 in its current implementaion can test this. The fonts have fvar instances for text and display families. The text instances uses opsz 17, while the display instances uses 18.

m4rc1e:
The github action to regression test the family already picks up this issue. When this PR is merged, we'll see clear diffs in the diffbrowsers images. I don't think I need to do anything here.

Imo, the transition from text to display is too long. If motion designers get their hands on this family, it is going to look weird. I mean look at this.

opsz_g

We can see that the transition happens from 17.9-18. Can we not make it say 17.99-18 or 17.98 if FF isn't happy? I'm guessing the level of granularity we have here is dictated by the fact the datatype is a F2DOT14.

RickyDaMa:
Yeah the whole point of the PR is to reduce the granularity of the transition, because otherwise MacOS Firefox ignores it entirely (per the original issue)

I don't know what the maximum value is we could use while still fixing the buggy behaviour, I just found that this one worked through trail & error. Based on the Firefox bug report, [this section](https://searchfox.org/mozilla-central/rev/ffa4d00965c5281def6d3ddcbcdf6259d38c9b9a/gfx/thebes/gfxMacFont.cpp#58-105) is likely the cause of the problem. If someone wanted to reverse engineer it and work out the best value, we could potentially have a tighter transition. The code doesn't mean much to me though :(

This having been said, the transition is still a relatively small part of the axis

m4rc1e:
> I don't know what the maximum value is we could use while still fixing the buggy behaviour, I just found that this one worked through trail & error. Based on the Firefox bug report, [this section](https://searchfox.org/mozilla-central/rev/ffa4d00965c5281def6d3ddcbcdf6259d38c9b9a/gfx/thebes/gfxMacFont.cpp#58-105) is likely the cause of the problem

Thank you very much for digging! In the code, there's `const float kOpszFudgeAmount = 0.01f;` so I'm guessing having `17.9` is the safest val we can have.

I'm happy with this PR. We can always make it tighter once the bug has solved.

m4rc1e:
Let me push a copy of this branch so it triggers diffenator2.

m4rc1e:
Sorry for my very late response to this PR. I needed to fix mac image diffing in order to test this.


It looks like the avar tweaking has worked. Here's the diff for Mac OS FireFox:


Regular-Italic-Bold-Bold Italic-Text Regular-Text Italic-Text Bold-Text Bold Italic-diffbrowsers_waterfall-macOS-12 7 2-x86_64-i386-64bit_firefox_120 0 1

We can see that the old style is text (<18 pt) and the new style is display (>= 18pt), which is what we want

If I look at Mac OS Chrome, we can see that both the old and new are display so FireFox now matches correctly.


You may be wondering why the style doesn't change in the diff image when the pt size increases. Diffenator2 uses the font's instance values so it has set them to 18.0 for the opsz axis for all sizes. 

Diffs can be found here: https://github.com/googlefonts/googlesans/actions/runs/7571450360

I'll take another look before I confirm this PR is good to merge.

m4rc1e (changes requested):
Just had a meeting with Marianna. We noticed upon closer inspection that the "G" isn't properly formed when the opsz axis is set to 18.0 in Mac FireFox.

Screenshot 2024-02-16 at 14 41 53

We then changed the opsz value to 100.0 and it still remained so something fishy is up.

MariannaPaszkowska:
Thanks for posting that, @m4rc1e! We will look into that

madig:
I played around with the numbers a bit to see if I can fix the "G" issue, and I found that even with `<map input="17.5" output="17.4"/>`, i.e. making a very slightly faster changeover than not having the bending at all, you don't hit the source opsz=18 in Firefox macOS at opsz=18. So, I don't think we can fix this without dropping the mapping tricks completely, unless there's a magic value somewhere. Or the bug in Firefox gets fixed.

PR #633, opened 2024-04-24
Update fontbakery and custom profile definitions

### LINKED ISSUE #316: Disable outline check on CI
https://github.com/googlefonts/googlesans/pull/315

Figure out where an outline check would be most useful and hook it in there.

belluzj:
Pending issue upstream: https://github.com/fonttools/fontbakery/issues/4671


RickyDaMa (approved):
Good to go, looks like some new checks are failing, but no issues with the Fontbakery update itself anymore

RickyDaMa:
Feedback raised about `com.google.fonts/check/case_mapping`: https://github.com/fonttools/fontbakery/issues/4733

FAILs here: https://github.com/googlefonts/googlesans/actions/runs/9157216680/job/25173357621#step:5:677

MariannaPaszkowska:
The FB check that is Failing is being updated here: https://github.com/fonttools/fontbakery/pull/4737

RickyDaMa:
The changes to that check have been merged, just need to wait for Fontbakery 0.12.7 now

RickyDaMa:
@MariannaPaszkowska the above PR you linked has just been released in a new version of Fontbakery, meaning this PR is now ready to merge. I've reverted the change that made `com.google.fonts/check/case_mapping` a WARN instead of FAIL now also

Hoolean:
Landing this PR would be helpful for incorporating the Font Bakery profile from the Pixel repo (#650); it's only a single check right now, but it would mean we don't have to backport it to the older version of Font Bakery on `main`, and then re-do the work in this PR to update it and any additional Android subset QA checks we may add later.

@MariannaPaszkowska when you have a mo', would you be able to review and merge? :)

Hoolean:
(in the meantime, I've pushed an empty commit so that we can make sure the checks are still passing with the latest Font Bakery v0.12.10 before merging)

Hoolean:
~We're hitting a timeout because the longer `interpolation_issues` check is running; we need to increase that timeout to accommodate it, but also fix the logic we added to make it only run on `main` and releases~

*Inaccurate, see below*

Hoolean:
Sorry, the above comment came from my misreading of GitHub action logs: the `interpolation_issues` check only runs on `main` and releases, as planned in #316.

---

The root cause of the timeout is that newer versions of Font Bakery take longer to run on Google Sans: previously our checks took 2 minutes for the variable fonts and 1 minute for statics, but we exceed our 5 minute cap for the variable checks now, and likely will for the statics too.

Newer versions of Font Bakery need more time, and run checks for all TTFs in a single process and CPU core, and so we have increased the timeout for checks to 15 minutes, but also moved to less expensive GitHub runners with fewer cores too.

Hoolean:
For reference, the last variable checks - newly on the smaller runner - took 5m55s. We expect this to be longer on `main` and for releases owing to the additional `interpolation_issues` check that runs on them, hence the need for a 15m timeout to avoid losing check runs to timeouts.

Hoolean:
@MariannaPaszkowska After the new fixes, ready for review and merge again!

Notable changes in PR:
- Updated Font Bakery to the latest version (at least, until the upcoming release)
- This includes Behdad's `interpolation_issues` check (configured to only run on `main` and releases)
- Increased allowed minute usage for GitHub Actions, and use smaller ones, now that Font Bakery takes longer

Extra notes:
- Merging will help us avoid extra unnecessary work on the Android subset QA setup (#650)

PR #647, opened 2024-09-04
Create ymetrics.py to eval glyph bounds vs. typo metrics bounds

chrissimpkins:
An initial script pass to define problem areas where glyph y-bounds extend beyond OS/2 typo bounds (#646)

This is meant to kick start the review in #646 and is based on the approach that Harry is proposing for Win metrics checks in https://github.com/googlefonts/googlesans-flex/pull/1010

I'll post a report from the current iteration of the script in #646.


scripts/ymetrics.py
```diff
@@ -0,0 +1,150 @@
...
+            uni_script = "Unknown"
+            uni_block = "Unknown"
+            uni_name = "Unknown"
+            uni_codepoint = "Unknown"
+            uni_category = "Unknown"
+            raise e
+
+        bad_flag = None
+
+        if y_min and y_max:
```
belluzj: add `is not None` because in Python, 0 is false

chrissimpkins:
Do we need this any longer?

belluzj:
We shouldn't need to run these scripts again in the future, but I think it would be good to archive them, maybe we could move them into a subfolder of `scripts/`?

madig:
I moved the script to its own subfolder. It can be merged like this.

PR #648, opened 2024-09-05
Add per-script subset list + subset operation in the pipeline

### LINKED ISSUE #646: [Android distribution issue] Glyph y-min and y-max values that extend beyond OS/2 typo metrics

### Phase 1 

- release and ship to Android

### Phase 2

- develop those families

belluzj:
Fixes #646 

- Adds an Android subset to the build/release pipeline which can be configured to exclude unwanted scripts
- Fixes GitHub workflows to allow for proper re-use of font artifacts between builds
- Adds draft script that can be used to ‘unmerge’ certain scripts from Google Sans into a per-script subfamily


scripts/one-off/make-subset-lists.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+def main():
+    font = Font.open("source/GoogleSans/GoogleSans-opsz18-wght380-GRAD0.ufo")
+    glyph_order = makeOfficialGlyphOrder(font)
+    cmap = makeUnicodeToGlyphNameMapping(font, glyph_order)
+    feature_file = parseLayoutFeatures(font)
+    gsub = compileGSUB(feature_file, glyph_order)
+    glyphs_by_script: dict[str, set[str]] = classifyGlyphs(
+        unicodedata_script, cmap, gsub
```
Hoolean: We may need to use `script_extension` here instead, let's review

belluzj: Done, thanks!

.github/workflows/cicd.yml
```diff
@@ -70,6 +70,30 @@ jobs:
             build/GoogleSans/variable
             build/GoogleSans/.intermediate
 
+  build-android:
+    runs-on: googlefonts-64cores-256GB
+    steps:
+      - name: Check out source repository
+        uses: actions/checkout@v3
+      - name: Set up Python v3.11 environment
```
RickyDaMa: 3.12 😉

.github/workflows/cicd.yml
```diff
@@ -70,6 +70,30 @@ jobs:
             build/GoogleSans/variable
             build/GoogleSans/.intermediate
 
+  build-android:
+    runs-on: googlefonts-64cores-256GB
```
RickyDaMa: Do we need this big of a running for two subsetting operations? Surely a free runner is fine here

RickyDaMa (commented):
Workflow looks good! Did we ought to make it conditional though, similar to building statics?

.github/workflows/cicd.yml
```diff
@@ -70,6 +70,30 @@ jobs:
             build/GoogleSans/variable
             build/GoogleSans/.intermediate
 
+  build-android:
```
belluzj, line 83: Missing a needs: variable-fonts

Hoolean:
Tasks to produce alpha files:
- [x] Produce Latin subset, remove Latin from other subsets
- [x] Make `name` tables unique enough for installing the fonts alongside each other
- [x] Apply per-subset typo metrics
- [x] Choose typo metric values for testing
- [x] Decide what to ship for Vietnamese initially
- [x] Decide which scripts can be merged into a single subset initially
- [x] (anything else?)

Hoolean:
I've now added the subset definitions for the alpha review files: the subsets are sourced from [here](https://docs.google.com/spreadsheets/d/1dLS5QQxldnsYbmBGmyDk9yYDFuZGrC1WBZSg51nPXXA/edit?usp=sharing), with their metrics sourced from [here](https://docs.google.com/spreadsheets/d/142QOCsodsEjcZ-mhbfV3DZM7g0xUpMoX5IOV1uaSiYc/edit?usp=sharing).

Hoolean:
Here are the TTFs currently being generated by this branch:

[AndroidSubsets_20240926_1616.zip](https://github.com/user-attachments/files/17151093/AndroidSubsets_20240926_1616.zip)

Hoolean:
New TODOs from meeting:
- [x] Change family name suffix
- [x] Change version number
- [x] Produce Vietnamese subset(s)
- [x] Recompute codepage ranges
- [ ] Why does the initial HTML report show extra space?
- [ ] Extend HTML report to show high/low/non-clipping AOSP words for each subset
- [x] Anything else for this first delivery in the brief?

Hoolean:
This is the (as yet, unexpanded) HTML report we showed in the meeting earlier to demonstrate the subsets being automatically selected and adjusting the line height in a browser environment:

Demonstration showing the overall line height increase when the browser selects one of the taller subsets

Place in the same folder as the subset fonts in the ZIP above to see locally: [index.html.zip](https://github.com/user-attachments/files/17155101/index.html.zip)

Hoolean:
New subset ZIPs!

[AndroidSubsets_20240927_2006.zip](https://github.com/user-attachments/files/17168577/AndroidSubsets_20240927_2006.zip)

These implement [the TODOs](https://github.com/googlefonts/googlesans/pull/648#issuecomment-2377737348) we identified together in our meeting, including: separating out previously-clipping Latin codepoints into a separate subset; adjusting the version number; adjusting the subset name suffixes; and some fixes to the subset process.

---

@chrissimpkins @MariannaPaszkowska

These will likely form the basis of the alpha files for the Android team to review, but to make sure the testing is as effective as possible we would first like to package them with a sample text document, as well as discuss together the nuances of the (now several) Latin variants we have produced.

The former we'll prepare on Monday, and then let's discuss the latter in our usual meet :)

chrissimpkins:
The prototype build version that Harry linked in https://github.com/googlefonts/googlesans/pull/648#issuecomment-2379908206 was distributed to Android SysUI eng for memory profiling.  This is the first stage of testing to approve this new approach on the Android system.

madig:
@belluzj recommends saving these two files from this PR and moving them over to the Fallback repo, then closing this PR:

- qa/samples/android_subsets.html
- scripts/gs_cross_subset_kerning.py

madig:
I have copied the files in https://github.com/googlefonts/pixel-brand-fallback-stack/commit/c178707f34de44890e5d1ebeb86b3fcff2ca1088.

PR #653, opened 2024-11-29
Update metadata

MariannaPaszkowska:
Adding an optimal set of subsets to ensure no glyphs that are defined in GF subsets are not subset out. 

Also, adding a few feature tags that need to be mentioned explicitly.


MariannaPaszkowska:
note to self: needs some updates, then @chrissimpkins will review.

MariannaPaszkowska:
note to self: symbols2 were not a needed subset

PR #655, opened 2025-01-16
Add sources/config.yaml

cmyr:
This will allow our sources to be found by fontc_crater.


chrissimpkins:
I believe that https://github.com/googlefonts/googlesans/pull/656 should fix our CI.  cairo is no longer available by default in recent releases of the GH Ubuntu runners.

cc @RickyDaMa @jackmcDaMa

PR #676, opened 2025-08-25
Avoid USE_MY_METRICS flags on components

### LINKED ISSUE #675: Avoid USE_MY_METRICS flags on components
Especially important when removing HVAR table from the fonts.

madig:
Closes https://github.com/googlefonts/googlesans/issues/675.


madig:
The shaping tests fail because fontbakery runs with a newer uharfbuzz, which seems to shape differently for some reason.

PR #680, opened 2025-09-23
Add BASE table using autobase

### LINKED ISSUE #646: [Android distribution issue] Glyph y-min and y-max values that extend beyond OS/2 typo metrics

### LINKED ISSUE #667: Add `BASE` table
The Android framework now respect min/max value in BASE table. 
Let's add a BASE table font for Vietnamese.

RickyDaMa:
Fixes #667 & #646


RickyDaMa:
Closing in favour of #683 & #684

PR #683, opened 2025-09-29
Add BASE table to family.fea (produced by `autobase`)

### LINKED ISSUE #646: [Android distribution issue] Glyph y-min and y-max values that extend beyond OS/2 typo metrics

### LINKED ISSUE #667: Add `BASE` table

RickyDaMa:
- Update requirements.txt so `fontTools` supports the BASE table
- Add autobase config file and plumb into Makefile
- Add fea code produced by autobase (edited, as currently it's not completely valid/complete, but the important metrics are there)
- Stop the subsetter from dropping the BASE table (upstream solution also being considered: https://github.com/fonttools/fonttools/pull/1866)


source/GoogleSans/family.fea
```diff
@@ -2024,3 +2024,49 @@ feature ss09 {
...
+    HorizAxis.MinMax kan2 dflt -787, 985;
+    HorizAxis.MinMax laoo dflt -474, 2537;
+    HorizAxis.MinMax latn dflt -292, 1204;
+    HorizAxis.MinMax latn VIT  -252, 1115;
+    HorizAxis.MinMax mlym dflt -395, 1068;
+    HorizAxis.MinMax orya dflt -1076, 1198;
+    HorizAxis.MinMax sinh dflt -362, 1092;
+    HorizAxis.MinMax tam2 dflt -482, 1036;
+    HorizAxis.MinMax tel2 dflt -1081, 1313;
+    HorizAxis.MinMax thai dflt -570, 1319;
```
m4rc1e: I'm very curious to see what the Android team have to say about adding every script. I avoided doing this in Roboto because I don't want to introduce any unforeseen regressions. I only added the requested Vietnamese entry.

Feel free to ignore this comment but I do think we need some testing from them before we can finalise each BASE table.

RickyDaMa: Yeah 100% not trying to say this is the best way or anything, just this is the output of `autobase` currently

RickyDaMa:
Note: there are changing to shaping tests, however when running the same words that are used in those tests through diffenator3, no visual difference was reported between the font with & without the BASE table

source/GoogleSans/family.fea
```diff
@@ -2048,25 +2048,23 @@ table BASE {
...
+    HorizAxis.MinMax grek dflt -286, 981;
     HorizAxis.MinMax guj2 dflt -648, 1244;
     HorizAxis.MinMax guru dflt -450, 1106;
-    HorizAxis.MinMax hebr dflt -469, 923;
+    HorizAxis.MinMax hebr dflt -469, 966;
     HorizAxis.MinMax khmr dflt -559, 1070;
     HorizAxis.MinMax kan2 dflt -787, 985;
     HorizAxis.MinMax laoo dflt -474, 2537;
     HorizAxis.MinMax latn dflt -292, 1204;
-    HorizAxis.MinMax latn VIT  -252, 1115;
```
RickyDaMa: Similar to Flex, Vietnamese is reporting more restrictive metrics than Latin, and is now presumably getting pruned because of it

MariannaPaszkowska: That doesn't look correct. I think we need better word lists.

RickyDaMa:
For those wondering, it looks like a new Fontbakery check is tripping up the CI, not anything related to the BASE table:

image

MariannaPaszkowska:
I reviewed the BASE table values for GS. So Gurmukhi and Lao seem to have a bit suspicious values. In Gurmukhi I still see clipping and Lao seem to have too generous top value (as per that text sample... we  need to investigate). Gurmukhi also seem ot have a shaping issue.

Lao:
Screenshot 2025-10-13 at 3 07 25 PM
Screenshot 2025-10-13 at 3 07 13 PM

Gurmukhi clipping and shaping:
@kalapi Could you tell us if Gurmukhi is shaping correctly here?  


Screenshot 2025-10-13 at 2 59 09 PM
Screenshot 2025-10-09 at 6 16 55 PM


Screenshot 2025-10-09 at 6 16 13 PM

RickyDaMa:
@MariannaPaszkowska done some rummaging into what we're seeing:

"from other" means "diffenator2". I've updated the script to reflect this, not that we'll be using it for much longer hopefully

Lao having really high metrics is from a specific word in the diffenator list. IIRC the Python script that generates the HTML report samples words randomly and so is perhaps not covering the most extreme word that gives the metrics: "ປົົົົົົົົກ" There are many mai kon stacked on top of each other, probably hence the height. Running `hb-shape` on the commandline confirms this. I expect other scripts where the exemplars in the report don't reach the full extent of the metrics are similar cases

As for why Gurmukhi isn't getting metrics as deep as it should, I'll need to investigate further. I can confirm the word given in `fontheight`'s word lists, and is part of the word list autobase uses to decide the metrics for that script

RickyDaMa:
Update on Gurmukhi: it appears that Harfbuzz (producing the reports) and Harfrust (powering autobase) shape some words differently, and this is why the BASE metrics are insufficient

Raised to harfrust people [here](https://chat.google.com/room/AAAAmakukgk/D-tmBleyoX4/D-tmBleyoX4?cls=10) (unable to reproduce with Noto, so providing a GS TTF privately)

MariannaPaszkowska:
You did well to update that script. We will still need to provide the script-specific subsets to address the clipping in Android in the legacy environments that don't support BASE.

Note to self: For Lao, we need to understand is this is really a rare and edge case " ປົົົົົົົົກ"... 

> @MariannaPaszkowska done some rummaging into what we're seeing:
> 
> "from other" means "diffenator2". I've updated the script to reflect this, not that we'll be using it for much longer hopefully
> 
> Lao having really high metrics is from a specific word in the diffenator list. IIRC the Python script that generates the HTML report samples words randomly and so is perhaps not covering the most extreme word that gives the metrics: "ປົົົົົົົົກ" There are many mai kon stacked on top of each other, probably hence the height. Running `hb-shape` on the commandline confirms this. I expect other scripts where the exemplars in the report don't reach the full extent of the metrics are similar cases
> 
> As for why Gurmukhi isn't getting metrics as deep as it should, I'll need to investigate further. I can confirm the word given in `fontheight`'s word lists, and is part of the word list autobase uses to decide the metrics for that script

RickyDaMa:
Updated the fea to the output of the latest autobase, which now uses LibreOffice + AOSP word lists, no longer using diffenator words.

There is a bug in its output, which I raised here: https://github.com/simoncozens/autobase/issues/10 (for now I substituted the `NULL` in `latn/VIT` for the corresponding value in `latn/dflt`)

For analysis, `fontheight` now has HTML reports!

```shell
# To install/update fontheight (needs a Rust toolchain installed & available)
cargo install --locked fontheight-cli
# To produce the report
fontheight "build/GoogleSans/variable/GoogleSans[GRAD,opsz,wght].ttf" --html --output GS.html
```

Here's a report I've generated: [~~GS.zip~~](https://github.com/user-attachments/files/23266389/GS.zip)

Update: smaller report with words deduplicated (see https://github.com/googlefonts/fontheight/issues/96): [~~GS.zip~~](https://github.com/user-attachments/files/23478462/GS.zip)

Update: outdated

RickyDaMa:
Here is the latest & greatest report, for both upright & italic Google Sans: 
[GS report new.zip](https://github.com/user-attachments/files/23726329/GS.report.new.zip)

Jany has made a number of improvements to the report, see https://github.com/googlefonts/fontheight/pull/97 for the full details, but in short the most extreme exemplars for each script (regardless of word list & VF location) are show at the top of each script heading, making it far harder to miss potential issues as they should be the first ones you see.

The good news is that, as expected, there's very little clipping. From my quick skim, nothing at all from AOSP/LibreOffice word lists. There are some example words from diffenator word lists that do clip, particularly for Gujarati & Gurmukhi, however per our discussion a couple weeks back, we're not considering diffenator word lists for BASE metrics due to lower trust in the source of the data, so this clipping on these words does not indicate any bugs/problems with fontheight/autobase, this is an intended choice on our part.

MariannaPaszkowska (approved):

PR #699, opened 2025-12-04
Add script to merge per-master Telugu features into single variable FEA

anthrotype:
I was trying to investigate the remaining ttx-diff when building GoogleSans.designspace with fontc vs fontmake and bumped into the fact that you are using multiple master-specific feature files and relying on fontmake building master GPOS tables and fontTools.varLib merging them into a variable GPOS + GDEF. This is currently used for some Telugu dist and blwm contextual positioning. 

This workflow is currently not supported by fontc (which is variable-first by design). But I need these diffs to go away so I could fix those that I need to fix. Since the per-master feature files are basically the same from a structural point of view except for the actual numbers, I figured that I could write a python script to turn them into a single variable feature file using the variable scalar syntax for value records and anchors that both fontc and fonttools feaLib already support.

To make them match I had to "massage" the telugu master feauture files a little bit (no functional difference, only minor things like order of glyphs inside a bunch classes, or a markClass that was merged in one file but split in another etc.).

I also run the merge script to produce two files (telugu-variable.fea and telugu-Italic-variable.fea) and modified the `include` statements in the main UFOs' features.fea so they point to those. The features are now identical for all the masters and this will trigger fontmake to build them "variably" instead of one master at a time + varLib.merge.

After this patch (in addition to disabling the ufo2ft `propagateAnchors` filter [which fontc doesn't support yet](https://github.com/googlefonts/fontc/issues/1704) but will eventually), when I compare GoogleSans fonts built with fontc vs fontmake via ttx-diff, I get _identical_ markkern.txt files (from otl-normalizer, with kerning and mark positioning lookups).

Feel free to accept or discard this. I made it mostly for fontc's own sake.


anthrotype:
@MariannaPaszkowska ping :)

MariannaPaszkowska:
@anthrotype many thanks! I will review asap.

MariannaPaszkowska (approved):

MariannaPaszkowska:
@anthrotype after merge. I found out that the `build-interpolatables` workflow is failing here: https://github.com/googlefonts/googlesans/actions/runs/20067458445/job/57559803664#step:4:158
Do you have time to look into that or should we take this on?

anthrotype:
oh.. I now see that job was skipped for my PR so I didn't notice.

right. So yeah you can't build `-o ttf-interpolatable` any more because now all the masters have (the same) variable feature file containing variable scalars and ufo2ft expects a static feature file for that kind of output.

But wait a second, why do you even need or care about building those interpolatable TTF masters? You should only care about variable font, if that compiles then it means the masters were indeed interpolatable...

belluzj:
@anthrotype @MariannaPaszkowska The interpolatable TTFs are needed for the [build workflow of Google Sans Arabic](https://github.com/googlefonts/googlesans-arabic/tree/master/sources/GoogleSans), which uses a TTF merge to mix the Latin glyph set into the Arabic fonts. See [the issue](https://github.com/googlefonts/googlesans/issues/471) and [this PR introducing these in the GS build](https://github.com/googlefonts/googlesans/pull/496).

In order to unblock Marianna to make a release of Google Sans, we thought about the following options:
* revert this PR (which is for fontc; but we don't use fontc yet here)
* keep this PR but adapt the workflow here to unmerge the feature files on the fly before building the interpolatable TTFs (basically do the opposite of what Cosimo did, make that part of the build workflow for interpolatable TTFs)
* keep this PR and implement support in ufo2ft for instantiating variable feature code into static feature code, so that ufo2ft can build interpolatable masters from this repo. See [this issue from Harry](https://github.com/googlefonts/ufo2ft/issues/945), and [this work started towards that goal](https://github.com/fonttools/fonttools/pull/3938). However this still requires some work to be done.

There's another option, which is to stop making those entirely, and change the build system of GS Arabic to do a UFO-level merge instead, like other GS script families. But that might be much toil for no benefit, so it's probably not a good idea.

What do you think is the best way to unblock the release now? And then to move forward in the future?

MariannaPaszkowska:
Because of the tight deadline I'd say:
1. Revert for now, so we are unblocked to release on Monday. 
2. And then work towards keeping this PR by implementing support in ufo2ft for instantiating variable feature code into static feature code, so that ufo2ft can build interpolatable masters from this repo. (Amazing that work on this started already! Thanks!)
CC @aaronbell

aaronbell:
I agree with Marianna's take to unblock and working towards keeping this PR via instantiation of variable feature code. We will be moving to fontc at some point, I'm sure, so all this will become necessary. 

The one other callout I wanted to make is that my preference is to have all aspects of a font family build using as generic / universal a process as possible. So to that extent, I'd like the GS Arabic to be updated at some point to follow the standard approach for this project (unless there's a compelling reason not to). What would be the size of that project? Or would it just be easier to eventually do a overhaul of the build system when we shift over to fontc and bring everything into alignment at that point?

RickyDaMa:
I've opened #707 for us to revert this PR while we wait on better upstream support

madig:
Can we revert the revert now that GS Arabic is merged into GS? Was GS Arabic the last consumer of the interpolatable builds?

anthrotype:
I am working exactly on that! hold the horses please!!

MariannaPaszkowska:
Many thanks @anthrotype you beat me to it @madig, I was just adding that on the task list!

PR #703, opened 2025-12-16
Add Riyal

### LINKED ISSUE #700: Add new Saudi Riyal symbol (U+20C1)
https://en.wikipedia.org/wiki/Saudi_riyal_sign

RickyDaMa:
Had to update some of our normalisation standards to try and mitigate all the noise coming from glyphsLib - I guess because it's been a while since we've merged work using this process Glyphs and/or glyphsLib have added new features, leading to new lib keys etc.

Shall keep this PR as a draft until I regression test it against `main` using diffenator


belluzj:
Notes about the normalization diffs:

* The kerning that seems to disappear was actually twice in kerning.plist; only one copy of it is kept
* The change of category from ligature to base for /vavvav-hb /vavyod-hb /yodyod-hb reflects this change in GlyphData.xml from 2023 that finally percolated all the way here: https://github.com/schriftgestalt/GlyphsInfo/commit/0a01300fb9c96c1810d7fd2dd06723889b9cc7d6#diff-dd6d8b538f22b839a2d5efa199d8dbb7105d10edb6f90bc940cee6670600b77aL4324-L4326

image

PR #529, opened 2022-09-13
Remove r-ligatures

MariannaPaszkowska:
Removed all references to `r` ligatures from `liga` and `dlig` features.

PR #713, opened 2026-01-14
Move the liga feature after smcp and c2sc

### LINKED ISSUE #712: Font defines ligature before smallcaps lookups
Found this while trying to make Android subsets. Fontspector says:

> OpenType small caps should be defined before ligature lookups to ensure proper functionality.
> 
> Rainer Erich Scheichelbauer (a.k.a. MekkaBlue) pointed out in a tweet (https://twitter.com/mekkablue/status/1297486769668132865) that the ordering of small caps and ligature lookups can lead to bad results such as the example he provided of the word "WAFFLES" in small caps, but with an unfortunate lowercase ffl ligature substitution.

It turns out GS has a section in family.fea's smcp block that turns ligatures back into smallcaps letters, to avoid the issue.

This didn't surface with the full fat font, because Armenian defines some smcp lookups first thing, so the check passes. The Android version strips out Armenian and its smcp lookups with it, triggering the check.

The fix would probably be simply to move the liga lookups near the end of family.fea and then drop the fixup section in the smcp block.

madig:
This moves the `liga` feature block after the `smcp` and `c2sc` blocks and removes the ligature decomposition hack in those (which still contained a dead r_f_f ligature decomposition that was left over from https://github.com/googlefonts/googlesans/pull/529). We can now enable the `smallcaps_before_ligatures` check again.

Closes https://github.com/googlefonts/googlesans/issues/712.

### DIFF
--- a/source/GoogleSans/family.fea
+++ b/source/GoogleSans/family.fea
@@ -387,34 +387,6 @@ feature case {
     lookup armn_case_contextual_substitution;
 } case;
 
-
-feature liga {
-    sub f f b by f_f_b;
-    sub f f h by f_f_h;
-    sub f f i by f_f_i;
-    sub f f j by f_f_j;
-    sub f f k by f_f_k;
-    sub f f l by f_f_l;
-    sub f f t by f_f_t;
-    sub f b by f_b;
-    sub f f by f_f;
-    sub f h by f_h;
-    sub f i by f_i;
-    sub f j by f_j;
-    sub f k by f_k;
-    sub f l by f_l;
-    sub f t by f_t;
-    sub t f by t_f;
-    sub t t by t_t;
-    sub [G g] o o g l e underscore l o g o by Google.logo;
-    sub [G g] o o g l e l o g o l i g a t u r e by Google.logo;
-    sub G l o g o l i g a t u r e by G.logo;
-    sub o l o g o l i g a t u r e by o.logo;
-    sub g l o g o l i g a t u r e by g.logo;
-    sub l l o g o l i g a t u r e by l.logo;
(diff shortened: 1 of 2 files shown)


source/GoogleSans/family.fea
```diff
@@ -1587,6 +1536,33 @@ feature smcp {
...
+    sub f t by f_t;
+    sub t f by t_f;
+    sub t t by t_t;
+    sub [G g] o o g l e underscore l o g o by Google.logo;
+    sub [G g] o o g l e l o g o l i g a t u r e by Google.logo;
+    sub G l o g o l i g a t u r e by G.logo;
+    sub o l o g o l i g a t u r e by o.logo;
+    sub g l o g o l i g a t u r e by g.logo;
+    sub l l o g o l i g a t u r e by l.logo;
+    sub e l o g o l i g a t u r e by e.logo;
```
khaledhosny, line 1563: Are small caps supposed to disable these logo ligatures? If not, they probably should be moved back before `smcp` feature.

MariannaPaszkowska: Thanks, Khaled, good catch! Yes, we should not touch brand ligatures, and therfore revert it. @madig Can we decompose the liga ligatures before applying smallcaps to solve the original issue?

khaledhosny: Composing then decomposing messes up clustering, which affects how output glyphs are mapped to input characters. Here is the font before this change. Notice how smallcaps fi is considered as one unit, even though they are two glyphs:
image


While with this change they are separate as expected:
image

My recommendation would to move the logo ligature to a separate lookup/feature block and have that only before smcp feature/lookup.

PR #720, opened 2026-02-19
Build derived font for Figma with 'Variable' suffix

Hoolean:
This is identical to the release font, other than family name and its derived name records.

All of the current variable font releases' instances are at the default `opsz` position, and so no further modification is required to accommodate Figma's auto-optical sizing behaviour.

## TODOs
- [x] Determine Font Bakery profiles
- [x] Test new workflow
- [x] Should the main release continue to remove non-default opsz instances?


MariannaPaszkowska:
@Hoolean Yes, we should continue to remove non-default opsz instances.

RickyDaMa:
In that case, this PR is ready to go!

RickyDaMa (approved):
Tested and working!

PR #729, opened 2026-04-07
Import Arabic from subfamily

RickyDaMa:
Mostly masterminded by the venerable @Hoolean, I just helped get it over the line.

Question for the original designer: are we okay to keep /mark-ar as it is in the pre-existing Google Sans source (that is, no outlines and zero width)? This keeps it in-line with the other control code characters in the font.

Outstanding work for a designer: the anchors on /dottedCircle need adding/repositioning from the Arabic sources, as the outlines were sufficiently different we didn't want to mechnically merge them.

TODO:
- [x] Update BASE table
- [x] Port Arabic's shaping tests: https://github.com/googlefonts/googlesans-arabic/tree/master/qa
- [x] Make Fontbakery happy
- [x] Whitespace widths
- [x] /dottedCircle anchors


MariannaPaszkowska:
@khaledhosny we are merging Arabic into GS, could you please have a look at Ricky's question?

khaledhosny:
> Question for the original designer: are we okay to keep /mark-ar as it is in the pre-existing Google Sans source (that is, no outlines and zero width)? This keeps it in-line with the other control code characters in the font.

The outlines and advance width should be ignored by OpenType layout engines (e.g. HarfBuzz will use `space` glyph, set its advance to zero, and ignore this glyph completely). The outlines get only used when control characters are preserved for debugging (e.g. `hb-view`’s `--preserve-default-ignorables`), so I prefer to have outlines for most control characters. But if this is the only one with outlines, I think it is fine to drop it for now.

khaledhosny:
Note you probably should not import the no-export glyphs like `beh-ar.init` (they all are no export and has the same color). They were there to help the designer visualize the dot placement, but they are otherwise not used. They also useful only with Glyphs automatic component alignment (as they would be updated when anchors changed), without that they likely to only serve as a source of confusion to future designers working on this.

scripts/one-off/arabic/import-arabic.py
```diff
@@ -0,0 +1,362 @@
...
+    "nine.tf",
+    "space",
+    "thinspace",
+    "hairspace",
+    "mark-ar",
+    "parenleft",
+    "parenright",
+    "question",
+    "radical",
+    "dottedCircle",
```
khaledhosny, line 76: Any difference in these glyphs outlines is unintentional (the first 3 are even marked no-export in the Arabic sources).

scripts/one-off/arabic/import-arabic.py
```diff
@@ -0,0 +1,362 @@
...
+    "zero.tf",
+    "one.tf",
+    "two.tf",
+    "three.tf",
+    "four.tf",
+    "five.tf",
+    "six.tf",
+    "seven.tf",
+    "eight.tf",
+    "nine.tf",
```
khaledhosny, line 67: Any differences here are unintentional as well, and are likely because we used earlier Latin source for these.

RickyDaMa:
Thanks for your comments Khaled, they're super helpful! I've culled all the excluded glyphs now 😌

RickyDaMa:
@khaledhosny hey Khaled, a couple more considerations we'd like your design eye on if you don't mind please:

1. /dottedCircle's outlines were quite different between Google Sans and GS Arabic, so we'd prefer if its Arabic-related anchors were re-positioned manually
2. Google Sans contains more whitespace glyphs than GS Arabic did. We currently use `locl` to reduce the width of /space /hairspace /thinspace to match the Arabic sources (see the `arabicspace_arabic` lookup), and we were wondering if it makes sense to also reduce the width of Google Sans' other whitespace glyphs in Arabic contexts. If you could either tell us the desired widths for the whitespace glyphs at the four master locations, or update the lookup yourself, that would be amazing!

image

khaledhosny:
> @khaledhosny hey Khaled, a couple more considerations we'd like your design eye on if you don't mind please:
> 
> 1. /dottedCircle's outlines were quite different between Google Sans and GS Arabic, so we'd prefer if its Arabic-related anchors were re-positioned manually

We have no specially requirement for dotted circle, as long as it has the relevant anchors it should be fine. Most probably GS didn’t have a dotted circle glyph, so we made our own.

> 2. Google Sans contains more whitespace glyphs than GS Arabic did. We currently use `locl` to reduce the width of /space /hairspace /thinspace to match the Arabic sources (see the `arabicspace_arabic` lookup), and we were wondering if it makes sense to also reduce the width of Google Sans' other whitespace glyphs in Arabic contexts. If you could either tell us the desired widths for the whitespace glyphs at the four master locations, or update the lookup yourself, that would be amazing!

Whitespace glyphs that are relative to the `space` width (like thin space or hair space) should be adjusted by the same percentage. But whitespaces that are somehow absolute, like emspace, enspace, figurespace, etc. I think should remain unchanged.

MariannaPaszkowska:
I suggest that for now we keep all spaces that have been designed for Arabic as localised versions, and we ignore other characters. I assume they were not a priority for Arabic, otherwise they would have been designed during the initial release. If we see interest in localising those other space characters, we can work on an update later on.
How does it sound?  @aaronbell @khaledhosny ?

aaronbell:
sgtm

khaledhosny:
Agreed.

PR #731, opened 2026-05-06
Update/Remove f-actions GitHub Actions

RickyDaMa:
Updates:
- f-actions/opentype-sanitizer
- f-actions/font-bakery

Removes:
- f-actions/font-setup
- f-actions/cldr-coverage (replaced by shapergloat)


MariannaPaszkowska:
note to self: we will use this PR once a couple of updates actions are released

PR #736, opened 2026-06-05
Update usWinAscent for Arabic

madig:
According to fontheight, the diffenator_arabic corpus has some Arabic words that can reach from -946 ("بُنيٍّ") to 1540 ("كُلٌّ") in some weights, so I'm 

1. adjusting the BASE metrics for Arabic, and
2. bumping usWinAscent to 1540 to avoid clipping.

Khaled suggested to also incorporate https://github.com/aliftype/quran-data/tree/main/quran into fontheight for testing, as it might surface higher heights and deeper depths.

### DIFF
--- a/source/GoogleSans/GoogleSans-opsz17-wght380-GRAD-50.ufo/fontinfo.plist
+++ b/source/GoogleSans/GoogleSans-opsz17-wght380-GRAD-50.ufo/fontinfo.plist
@@ -53,7 +53,7 @@
     <key>openTypeOS2VendorID</key>
     <string>GOOG</string>
     <key>openTypeOS2WinAscent</key>
-    <integer>1540</integer>
+    <integer>1509</integer>
     <key>openTypeOS2WinDescent</key>
     <integer>1079</integer>
     <key>postscriptBlueValues</key>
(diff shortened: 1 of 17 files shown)


RickyDaMa:
If we trust this word in the diffenator Arabic corpus that'll be annoying because currently autobase (which is what I use to generate the BASE table for the Google brand families) uses AOSP corpora due to lower trust in diffenator, which evidently gives lower metrics (else we wouldn't have this PR/discussion).

RickyDaMa:
I also don't think we should be increasing the vertical metrics beyond what `autobase` is advising, as that's the sanctified tool we have for the job. If we want to increase it beyond that, then that needs to come from autobase one way or another.

RickyDaMa:
I have opened https://github.com/googlefonts/fontheight/issues/101 about the use of the Quran as an Arabic word list

RickyDaMa:
1462 was chosen because it is 1 more than the head.yMax of the VF, to appease the Fontbakery check. Given some of the statics having a greater head.yMax than the VF, I would suggest choosing a value 1 more than the highest, which is 1509 (given Text Bold has the highest yMax at 1508). This appears to be how the previous value was chosen too.

PR #740, opened 2026-06-10
Add script to merge_variable_fea.py and apply to Telugu and Arabic features

### LINKED ISSUE #708: Re-add Telugu variable feature merging
"We will re-add this [Telugu variable feature merging] in future when ufo2ft supports instantiating variable feature code to static feature code, as detailed https://github.com/googlefonts/googlesans/pull/699#issuecomment-3677694648"

Related: 
https://github.com/googlefonts/googlesans/pull/699
https://github.com/googlefonts/googlesans/pull/707

anthrotype:
The script merges per-master FEA feature fragments into a single variable feature file by working on the feaLib AST rather than manipualting raw text wit regex (see the old reverted #699).
It walks the parallel statement trees together, performs a cross-master alignment check that zeros out values during comparison to catch structure mismatches loudly, and converts any ValueRecord fields or Anchor coordinates that vary across masters into VariableScalars.

In commit bbd67c672e08b72c5338df62660d78317646e47c, I also applied the merge script to the existing Telugo and Arabic features.

This allows to compile GoogleSans with fontc, which can't currently merge per-master features like varLib does (https://github.com/googlefonts/fontc/issues/1829). Right now fontc silently ignores GS's per-master features (silently takes the first one) because its identical-features check is too naive (https://github.com/googlefonts/fontc/pull/2027 fixes that, but breaks fontc CI's smoke test where we build GS from `HEAD`).

It also continues to build with fontmake of course. I verified locally that compiling GS with these merged variable features produces exactly the OTL tables (after otl-normalizer) as building GS from `main` branch with individual per-master features.

Closes #708

### DIFF
--- a/scripts/merge_variable_fea.py
+++ b/scripts/merge_variable_fea.py
@@ -103,8 +103,10 @@ DEVICE_SLOTS = ("xPlaDevice", "yPlaDevice", "xAdvDevice", "yAdvDevice")
 
 _GLYPH_CLASS_RE = re.compile(r"\[([^\]]+)\]")
 
+Location = dict[str, int | float]
 
-def parse_location(spec: str) -> dict:
+
+def parse_location(spec: str) -> Location:
     """'opsz=18,wght=400' -> {'opsz': 18, 'wght': 400} (int where integral)."""
     loc = {}
     for part in spec.split(","):
@@ -116,13 +118,13 @@ def parse_location(spec: str) -> dict:
     return loc
 
 
-def glyph_names_from_ufo(ufo_path: str) -> set:
+def glyph_names_from_ufo(ufo_path: str) -> set[str]:
     return set(UFOReader(ufo_path).getGlyphSet().keys())
 
 
 def parse_master(
-    frag_path: str, glyph_names: set, context: list, include_dir: str
-) -> list:
+    frag_path: str, glyph_names: set[str], context: list[str], include_dir: str
+) -> list[ast.Statement]:
     """Parse one fragment with prepended context includes; return only the
     fragment's own top-level statements (context statements sliced off)."""
     preamble = "".join(f"include({c});\n" for c in context)
@@ -140,7 +142,7 @@ def parse_master(
     return full.statements[n_ctx:]
 
 
-def meaningful(stmts: list) -> list:
+def meaningful(stmts: list[ast.Statement]) -> list[ast.Statement]:
     """Statements that carry structure/values (standalone comments dropped from
     pairing; they remain in the reference tree and so in the output)."""
     return [s for s in stmts if not isinstance(s, ast.Comment)]
@@ -150,7 +152,7 @@ def is_block(node) -> bool:
     return hasattr(node, "statements") and isinstance(node.statements, list)
 
 
-def block_header(node) -> tuple:
+def block_header(node) -> tuple[str, str | None, bool | None]:
     """Block identity independent of its body (and its body's comments)."""
     return (
         type(node).__name__,
@@ -200,20 +202,20 @@ def signature(stmt) -> str:
     )
 
 
-def shape(node) -> tuple:
+def shape(node) -> tuple[str, str | None, bool | None] | tuple[str, str]:
     """Per-statement alignment key: block header, or leaf signature."""
     if is_block(node):
         return block_header(node)
     return ("leaf", signature(node))
 
 
-def aligned(mlists: list) -> bool:
+def aligned(mlists: list[list[ast.Statement]]) -> bool:
     """True if every master has the same statement shapes in the same order."""
     shapes = [tuple(shape(s) for s in ml) for ml in mlists]
     return all(s == shapes[0] for s in shapes)
 
 
-def value_holders(stmt) -> list:
+def value_holders(stmt) -> list[ast.Anchor | ast.ValueRecord]:
     """Ordered Anchor/ValueRecord objects carrying mergeable numerics in `stmt`
     itself (not its child statements). [] for purely structural statements;
     ligature-caret positions are bare numbers, not holder objects, and are merged
@@ -237,7 +239,7 @@ def value_holders(stmt) -> list:
     return []
 
 
-def holder_slots(h) -> tuple:
+def holder_slots(h) -> tuple[str, ...]:
     """Numeric slot names on a value holder; raises on shapes the merge can't
     express (device tables, contourpoint anchors). NULL anchors never reach here:
     feaLib parses '<anchor NULL>' to a value-less component, so value_holders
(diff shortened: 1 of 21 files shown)


anthrotype:
> I verified locally that compiling GS with these merged variable features produces exactly the OTL tables (after otl-normalizer) as building GS from main branch with individual per-master features.

Note, this comparison surfaced a real bug in ufo2ft variable kernFeatureWriter which is/will be fixed with https://github.com/googlefonts/ufo2ft/pull/990

(fontc still has the exact same bug and I will be fixing it next)

scripts/merge_variable_fea.py
```diff
@@ -0,0 +1,528 @@
...
+    for part in spec.split(","):
+        tag, sep, val = part.partition("=")
+        if not sep:
+            raise ValueError(f"bad location component {part!r} in {spec!r}")
+        f = float(val)
+        loc[tag.strip()] = int(f) if f.is_integer() else f
+    return loc
+
+
+def glyph_names_from_ufo(ufo_path: str) -> set:
```
madig: `-> set[str]` to be precise, likewise for other type annotations.

madig:
Did a trace through the code and used it in an internal project and it worked there.

anthrotype:
> this comparison surfaced a real bug in ufo2ft variable kernFeatureWriter... https://github.com/googlefonts/ufo2ft/pull/990
> (fontc still has the exact same bug... https://github.com/googlefonts/fontc/pull/2040

both compilers fixed

MariannaPaszkowska:
@madig @anthrotype is this ready for a review/merge?

anthrotype:
I think so. As I said above, I verified that

> compiling GS with these merged variable features produces exactly the OTL tables (after otl-normalizer) as building GS from main branch with individual per-master features

but @madig or others may wish to reproduce this and confirm on their end too

anthrotype:
you also need the latest https://github.com/googlefonts/ufo2ft/releases/tag/v3.8.2 released a couple weeks ago containing the kern writer bug fix that GS hits

madig:
After updating all requirements, including ufo2ft, I indeed get identical mark/kern output from otl-normalize when comparing the upright VF from main and this one, but there's a whole bunch of GSUB diffs I can't quite place. Many of them seem to be reorderings, but some look just different.

anthrotype:
I would not expect GSUB diffs.. What GS sources and ufo2ft versions are you comparing exactly?

anthrotype:
> comparing the upright VF from main and this one

that's the wrong comparison. You should be comparing the source from main vs the source from this `merge-fea-ast` branch, built with the same (latest) ufo2ft version (which should probably pull in the latest/current of everything, fontmake, fonttools whatever), if you _only_ want to compare the effect of building from different per-master features.fea vs one single merged variable features.fea.
If you mix too many variables it's gets hard to pin down.

madig:
What I did:

1. Check out main and this branch in separate dirs
2. Update the requirements.txt file for main and merge-fea-ast to the same state, by simply updating all deps including ufo2ft 3.8.2
3. Build the VFs for both branches, then run `uvx ttx-diff --normalizer_path /path/to/otl-normalizer --outdir /tmp --fontmake_font build/GoogleSans/variable/GoogleSans\[GRAD,opsz,wght\].ttf --fontc_font ../googlesans-varfea/build/GoogleSans/variable/GoogleSans\[GRAD,opsz,wght\].ttf`. Some tables have inconsequential diffs, but GSUB has loads.

anthrotype:
> uvx ttx-diff

no, the version of ttx-diff on pypi is probably outdated, when you compare fontc/otl-normalizer output you must use the same version that is is in the fontc repository, so install in editable mode from the the latest fontc local checkout

madig:
Running ttx-diff from fontc commit b4618b5aacd641f746276368c0180a61ad866259 like so:

```
cd fontc/ttx_diff
uv run ttx-diff --normalizer_path ~/fontc/target/release/otl-normalizer --outdir /tmp/aaa --fontmake_font ~/googlesans/build/GoogleSans/variable/GoogleSans\[GRAD,opsz,wght\].ttf --fontc_font ~/googlesans-varfea/build/GoogleSans/variable/GoogleSans\[GRAD,opsz,wght\].ttf
```

still shows up GSUB diffs.

Am I holding it wrong?

anthrotype:
> Am I holding it wrong?

I'll take a look

anthrotype:
oh wait, is the GSUB diff between fontmake and fontc fonts the _same_ in both main and the merge-fea-ast? If if's pre-existing, as I suspect it should be, I don't think it should block the current PR, which only ever touches GDEF/GPOS, never GSUB.

anthrotype:
why are you passing --fontmake_font and --fontc_font though? Should you not let ttx-diff build both?

madig:
Maybe I misunderstood how to compare the fonts? The GSUB diff is the difference between building the font on main with updated deps and merge-fea-ast with updated deps via `uv run make gs-vf`. This is supposed to test the impact of merging features vs. using a merged feature file. Note that I'm using the ttx-diff fontmake/fontc args to test the fonts built by the build system across branches, not to test fontmake vs. fontc!

anthrotype:
ok let's keep fontc out of the comparison, sorry for bringing that up and confusing things even more.

What you need to compare to reassure that this branch does the right thing is build the sources with fontmake only on main and on the `merge-fea-ast` branch using the exact same toolchain (with the latest ufo2ft).
Then you run otl-normalizer on those two fontmake fonts and see that the output from otl-normalizer is identical. Note that only GDEF and GPOS are handled by otl-normalizer, GSUB is not normalized by otl-normalizer. I expect the GSUB tables to be identical between fontmake(main) and fontmake(merge-fea-ast).

The way you're using ttx-diff to compare two fontmake fonts (rather than a fontmake font vs a fontc font) is not really supported at the moment, because ttx-diff hard-codes some logic to reduce the diff noise and applies that conditionally based on the compiler (supposedly) used to build the font.

You should expect to get that same GSUB diff if you passed the _same_ font to those `--fontmake_font`/`--fontc_font` CLI flags, which would be absurd. The diffs are spurious/introduced by ttx-diff tool itself in its compiler-specific (benign) normalizations. Have you tried?

anthrotype:
> Note that I'm using the ttx-diff fontmake/fontc args to test the fonts built by the build system across branches, not to test fontmake vs. fontc!

yes sorry my bad!

madig:
Ah! Okay. Doing a manual otl-normalizer run and a plain TTX diff shows only name IDs being different between the fonts, including in GSUB. Green light from me now :)

anthrotype:
> TTX diff shows only name IDs being different between the fonts, including in GSUB

as long as the strings are the same and the difference is only in the user-defined area > 256 then it should be fine. Want me to chase those down?

madig:
I didn't check the name table itself, but instead looked at the `<!-- Names -->` that fontTools spells out in TTX dumps and they are all the same in fvar, GSUB and STAT (the tables aside from GPOS that differ), only the numbers change, all above 256. I don't think it's worth chasing them down.

madig:
I just saw that in the var fea font, one duplicated "Regular" name table entry is dropped, as I suspected.

madig (approved):

MariannaPaszkowska:
Thank you all! 🚀  This will be very helpful for us as we move towards fontc.

PR #760, opened 2026-07-21
Set release fonts to use post v3

### LINKED ISSUE #115: turn post into a format 3 table?
I just saw that the `post` table takes up 28KB in the statics and VFs. Is there a hard need for it?

RickyDaMa:
During the final release step, all fonts are edited to use post v3, granting savings in file size documented in #115 & #757 

This approach is preferable to #757 because it doesn't break our QA workflows

Did we ought to integrate this into the Makefile somehow? I'm hesitant to overwrite/modify either the intermediate or variable folder fonts because then that'll break QA (again), but also creating a new folder/package for these seems overblown and cumbersome.

### DIFF
--- a/.github/workflows/cicd.yml
+++ b/.github/workflows/cicd.yml
@@ -251,6 +251,13 @@ jobs:
     permissions:
       contents: write
     steps:
+      - name: Check out source repository
+        uses: actions/checkout@v7
+      - name: Set up Python v3.12 environment
+        uses: actions/setup-python@v6
+        with:
+          python-version: '3.12'
+          cache: 'pip'
       - name: Save tag name
         id: vars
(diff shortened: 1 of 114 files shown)


Hoolean:
I'd prefer it in the Makefile in some form, so that we get the same fonts when building locally, and to keep the build steps all in one place. I'm less sure about whether it should overwrite or modify in place, though...

RickyDaMa:
Yeah I'm just also conscious of the more edits I make (ha) to the Makefile that's just effort that's going to be overwritten as #753 progresses

MariannaPaszkowska:
I agree with @RickyDaMa that considering the changes we want to introduce soon, we shouldn't spent too much time on that.

PR #772, opened 2026-08-28
Update requirements & migrate to fontc 🚀

RickyDaMa:
Notable changes:
- Upstream, anchor propagation changed to consolidation algorithms across fontmake, fontc, and Glyphs. This resulted in regressions in Devanagari for us. To mitigate this, I ran the old `PropagateAnchors` filter on the sources, and committed the added anchors. I then specified the filter chain manually in the designspaces instead of relying upon defaults, to exclude the new anchor propagation logic for being run.
- Source changes made in #774 
- fontc inserts its own version into the name ID 5 version string

~~End result: zero pixel diffs~~

~~Diffenator report, comparing #768 (gftools + fontmake) to gftools + fontc (warning: anticlimactic): [fontc-gftools.zip](https://github.com/user-attachments/files/31555436/fontc-gftools.zip)~~


anthrotype:
cool!

> This resulted in regressions in Devanagari for us

as in actual bugs or just mere diffs? could you have simply added explicit anchors only for those and let the rest propagate normally so you would avoid having to override the default filter list?

anthrotype:
> I then specified the filter chain manually in the designspaces instead of relying upon defaults, to exclude the new anchor propagation logic for being run.

wait, for UFO we don't automatically propagate anchors unless propagateAnchors filter appears in the lib. So if you don't want that, just omit the whole `com.github.googlei18n.ufo2ft.filters` and let the default ones run.
For Glyphs, we do apply by default and there's a `Propagate Anchors` custom parameter to disable.
There's also a tri-state CLI flag `--propagate-anchors` which you can set to true|false (defaults to whatever the source format defaults)

RickyDaMa:
> as in actual bugs or just mere diffs?

I'm no designer, but it was definitely undesirable behaviour. Here's a diffenator report: [deva-regression.zip](https://github.com/user-attachments/files/31556040/deva-regression.zip)


> could you have simply added explicit anchors only for those and let the rest propagate normally?

Annoyingly, even after propagating the anchors manually, keeping the default filter chain resulted in the regressions still

Separately, setting fontc-specific flags when we're using fontc through gftools is an added challenge, which I've opened a PR to ease: https://github.com/googlefonts/gftools/pull/1219

anthrotype:
> I'm no designer, but it was definitely undesirable behaviour

maybe it's related to this? https://github.com/googlefonts/fontc/issues/2007

> I then specified the filter chain manually in the designspaces

I see you need to override that because you pass filters that aren't on by default like flattenComponents and decomposeTransformedComponents, you can't just omit the key entirely because you intentionally don't want just the default ones.
Omitting the propagateAnchor filter in the DS lib overrides whatever was in the individual UFO libs and it works in both compilers (if the purpose is to disable it in the source), however you may also want to delete the propagateAnchors filter that is still present in the master UFOs' lib.plist

anthrotype:
> Here's a diffenator report: [deva-regression.zip](https://github.com/user-attachments/files/31556040/deva-regression.zip)

oh! I see what's going on. The glyphs involved in the diff contain anchors with contradictory roles that trigger a cancellation step in the propagation algorithm: if the glyph has `_bottom` it drops `top` and `_top`, if it has `_top` it drops `bottom` and `_bottom`. From Glyphs.app PoV you can't both say "I am a mark that attach below a base" and "I am a base and marks may attach above me" (or viceversa).

e.g. `da_uMatra-deva` and `da_uuMatra-deva` have `_bottom` alongside `top`, `nga-deva` and `ng-deva` have `_top` as well as `bottom`.

See fontc propagate_anchors.rs (there's equivalent code in glyphsLib and ufo2ft since all three now follow the same algo):
https://github.com/googlefonts/fontc/blob/6673d77d7d1878e77652542dddf7192adf9ebbba/fontir/src/propagate_anchors.rs#L411-L413
https://github.com/googlefonts/fontc/blob/6673d77d7d1878e77652542dddf7192adf9ebbba/fontir/src/propagate_anchors.rs#L453-L461

The step runs after the glyph's own anchors have been merged in so it ends up deleting human-authored anchors, not just propagated ones (maybe something to fix upstream).

In any case, those specific glyphs are declared as GDEF bases so their "_"-prefixed anchors are never used for attachment in the compiled font. If you remove them, you can stop that cancellation from triggering and you can keep the filter running without having to bake all the anchors in the source.

I can PR this if you want?

RickyDaMa:
> In any case, those specific glyphs are declared as GDEF bases so their "_"-prefixed anchors are never used for attachment in the compiled font. If you remove them, you can stop that cancellation from triggering and you can keep the filter running without having to bake all the anchors in the source.
> 
> I can PR this if you want?

Thanks so much for digging into this! If you're able to, that would be great thanks 😊 If you base your PR on #768 or `main`, then I'll rebase this one on top (removing the obsolete commits)!

anthrotype:
@RickyDaMa see #774 

I based it off `gftools` branch (#768). Once that is in, this #772 can rebase on top and drop both the "Run old propagation" (11ad6602) and the designspace filter override (f57ac2c8) and propagation can keep running.

PR #774, opened 2026-08-28
Drop stray _bottom/_top anchors from four Devanagari composites

anthrotype:
See https://github.com/googlefonts/googlesans/pull/772#issuecomment-5454002344

`da_uMatra-deva` and `da_uuMatra-deva` have `_bottom` alongside `top`, `nga-deva` and `ng-deva` have `_top` alongside `bottom`, and the cancellation step in anchor propagation drops the conflicting `top`/`bottom`, leaving marks unattached.

But the `_` anchors do nothing (all four glyphs are GDEF bases, so nothing attaches through them) so removing them stops the cancellation and lets propagation keep running.

I built both designspaces with fontc 1.0.0-rc.2: the seven regressed Devanagari words now shape with the same mark offsets as before the propagation change. The only GPOS difference is on those four bases, and it's all additions.

A diffenator3 run against #772 as it stands shows no Devanagari differences.

Only two words differ elsewhere: Gujarati રૅયુનિઑઁ, where propagation gives the composite `oCandra-gujarati` a top anchor so `candraBindu-gujarati` attaches (it doesn't in #772); and Hebrew װָאךְ, where the vowel point lands a few units off. I haven't chased the latter down. Maybe that's worth a designer's eye.

### DIFF
--- a/sources/GoogleSans-opsz17-wght380-GRAD-50.ufo/glyphs/da_uM_atra-deva.glif
+++ b/sources/GoogleSans-opsz17-wght380-GRAD-50.ufo/glyphs/da_uM_atra-deva.glif
@@ -1,7 +1,6 @@
 <?xml version='1.0' encoding='UTF-8'?>
 <glyph name="da_uMatra-deva" format="2">
   <advance width="600"/>
-  <anchor x="457" y="-31" name="_bottom"/>
   <anchor x="414" y="679" name="top"/>
   <outline>
     <contour>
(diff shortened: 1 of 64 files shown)


RickyDaMa:
Can confirm diffenator flags this

Before:
image

After:

image

Will review with Marianna this morning

MariannaPaszkowska:
@RickyDaMa @anthrotype 
@kalapi confirmed that this is a regression... 
So we need to investigate this further.

anthrotype:
Sorry which one is the buggy rendering? I presume the "After" one where the mark is detached.. And what produced which (fontmake vs fontc)?
Can you tell me the exact strings?

anthrotype:
> And what produced which (fontmake vs fontc)? Can you tell me the exact strings?

sorry I was typing from my phone and I'm context switching a lot lately.. Replying to myself: the string in those screenshot is same one I flagged last week (from the PR body above):

> Gujarati રૅયુનિઑઁ, where propagation gives the composite oCandra-gujarati a top anchor so candraBindu-gujarati attaches (it doesn't in https://github.com/googlefonts/googlesans/pull/772)

Is this specific attachment the unwanted one that produces the regression? that means "Before" (with the attachment) is buggy and "After" (detached) is the expected.

